PR #7, opened 2021-11-05
Build and test runtime on Windows

pcolberg:
Use actions cache to reduce aocl setup time from between 1 to 10 min for
downloading and running the installer to 20 sec for restoring the cache.

Run on Windows 2019 only since Windows 2016 hosted runners are
deprecated and will be removed from GitHub actions on 2022-03-15.


zibaiwan (approved):
Thanks Peter!

PR #10, opened 2021-11-09
Add Klocwork scanning workflows

pcolberg:
The main branch scan obtains all Klocwork issues of the main branch and
uploads the results to a Klocwork server. This workflow requires access
to a Klocwork token to authenticate to the server for build and upload.

The pull request workflow does a full analysis of the base ref and an
incremental analysis of the head ref of the pull request using local
scans, which do not require a Klocwork server. This workflow fails if
the incremental analysis finds new issues introduced by the head ref.

### DIFF
--- a/.github/workflows/klocwork-main.yml
+++ b/.github/workflows/klocwork-main.yml
@@ -1,0 +1,75 @@
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+
+name: Klocwork main branch scan
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '**'
+      - '!**.md'
+      - '!**/.clang-format'
+      - '!**/COPYING'
+      - '!**/LICENSE'
+      - '!.github/**'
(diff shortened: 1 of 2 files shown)


pcolberg:
This is ready for review. For the pull request workflow, I initially experimented with [`actions/cache`](https://github.com/actions/cache) for caching the Klocwork analysis of the base ref, since each full analysis takes a bit less than 5 minutes, but the workflow seemed too complex and nontrivial to maintain. Then I dropped the `--rebuild` option to `kwcheck run`, which reduces the incremental analysis of the head ref to a minute. Combined with running ninja on multiple cores under `kwinject`, the result is a robust, straight-forward workflow that takes a bit more than 5 minutes altogether.

pcolberg:
Tested a [failing pull request workflow](https://github.com/intel/fpga-runtime-for-opencl/runs/4184163146?check_suite_focus=true) with an artificially induced issue, and a [passing main branch workflow](https://github.com/pcolberg/fpga-runtime-for-opencl/runs/4184167082?check_suite_focus=true) accessing the Klocwork token from environment secrets.

zibaiwan (approved):
Thanks Peter for your great work on this. The flow looks good to me!

sherry-yuan (approved):
Thank you @pcolberg, Looks good to me too.

PR #13, opened 2021-11-15
Limit container image push to current tag

pcolberg:
When using a self-hosted actions runner, `docker image push --all-tags`
also pushes all image tags built in previous container workflow runs.


zibaiwan (approved):
Thanks Peter.

PR #16, opened 2021-11-16
Add BDBA scanning workflow

pcolberg:
See [intel/fpga-runtime-for-opencl/runs/4231767557](https://github.com/intel/fpga-runtime-for-opencl/runs/4231767557?check_suite_focus=true) for a partial test run up before submission, which fails as expected since pull requests from forked repositories do not have access to secrets. This workflow will only run on the `main` branch.

PR #17, opened 2021-11-16
Add delta coverage check on pull requests

sherry-yuan:
[Failing Check](https://github.com/intel/fpga-runtime-for-opencl/runs/4242001567?check_suite_focus=true) by deleting acl_device_test

[slight coverage decrease](https://github.com/intel/fpga-runtime-for-opencl/runs/4255873986?check_suite_focus=true)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,81 @@
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+
+name: delta valgrind checks
+
+permissions:
+  # Grant read permissions to repository in case it is not a forked public
+  # repository, but a private repository that was created manually.
+  contents: read
+
+  # Grant read permissions to private container images.
+  packages: read
+
+on:
+  pull_request:
+    paths:
+      - '**'
+      - '!**.md'
+      - '!**/.clang-format'
+      - '!**/COPYING'
+      - '!**/LICENSE'
+      - '!.github/**'
+      - '.github/workflows/coverage.yml'
+      - '!.gitignore'
+      - '!cmake/manifests/**'
+      - 'cmake/manifests/linux/**'
+      - '!container/**'
+      - '!docs/**'
+      - '!scripts/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    container:
+      image: ghcr.io/intel/fpga-runtime-for-opencl/ubuntu-20.04-dev:main
+
+    env:
+      TMP_DIR: /home/build
+
+    steps:
+      - name: change ownership of workspace to current user
+        run: sudo chown -R build:build .
+      
+      - name: checkout main branch
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.base_ref }}
+      
+      - name: create build directories
+        run: mkdir "$TMP_DIR/child_build" "$TMP_DIR/parent_build"
+
+      - name: create parent build files
+        run: cd "$TMP_DIR/parent_build" && cmake -G Ninja "$GITHUB_WORKSPACE" -DCMAKE_BUILD_TYPE=Debug -DACL_CODE_COVERAGE=ON
+
+      - name: build parent runtime
+        run: cd "$TMP_DIR/parent_build" && ninja -v -k0
+
+      - name: parent runtime coverage scan
+        run: cd "$TMP_DIR/parent_build" && ctest -T Test -T Coverage
+
+      - name: checkout current branch
+        uses: actions/checkout@v2
+
+      - name: create child build files
+        run: cd "$TMP_DIR/child_build" && cmake -G Ninja "$GITHUB_WORKSPACE" -DCMAKE_BUILD_TYPE=Debug -DACL_CODE_COVERAGE=ON
+
+      - name: build child runtime
+        run: cd "$TMP_DIR/child_build" && ninja -v -k0
+
+      - name: child runtime coverage scan
+        run: cd "$TMP_DIR/child_build" && ctest -T Test -T Coverage
+      
+      - name: coverage status
+        run: ./scripts/coverage_diff.py "$TMP_DIR/child_build" "$TMP_DIR/parent_build"
+
+      - name: revert ownership of workspace to root
+        run: sudo chown -R root:root .
+        if: always()
--- a/scripts/coverage_diff.py
+++ b/scripts/coverage_diff.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Example usage: ./coverage_diff.py child_build_dir parent_build_dir
+import xml.etree.ElementTree as ET
+import sys
+import argparse
+import glob
+
+
+def want_file(fn):
+    return fn.startswith("./src")
+
+def txt_table(matrix):
+    s = [[str(e) for e in row] for row in matrix]
+    lens = [max(map(len, col))+1 for col in zip(*s)]
+    fmt = '\t'.join('{{:{}}}'.format(x) for x in lens)
+    table = [fmt.format(*row) for row in s]
+    return '\n'.join(table)
+
+def markdown_table(array):
+    """ Input: Python list with rows of table as lists
+               First element as header. 
+        Output: String to put into a .md file 
+        
+    Ex Input: 
+        [["File Name", "Coverage (%)", "Delta Coverage"],
+         ["filename1",  80           ,  10             ],
+         ["filename2",  70           ,  -5             ]] 
+    """
+
+    markdown = "\n" + "| "
+
+    for e in array[0]:
+        markdown += " " + str(e) + " |"
+    markdown += "\n"
+
+    markdown += '|'
+    for i in range(len(array[0])):
+        markdown += "-------------- | "
+    markdown += "\n"
+
+    for entry in array[1:]:
+        markdown += "| "
+        for e in entry:
+            markdown += str(e) + " | "
+        markdown += "\n"
+
+    return markdown + "\n"
+
+def parse_coverage(child_data, parent_data):
+    """
+    child_data: xml file string of coverage scan result for child build
+    parent_data: xml file string of coverage scan result for parent build
+    output: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
+    """
+    coverage = {}
+    # Parse parent coverage
+    root = ET.fromstring(parent_data)
+    for elm in root.findall(".//File[@FullPath]"):
+        fn = elm.attrib["FullPath"]
+        if not want_file(fn):
+            continue
+        percent_cov = float(elm.find('PercentCoverage').text)
+        if fn not in coverage:
+            # When file exist in parent, the file is not new
+            coverage[fn] = {'child_cov': 0, 'new_file': False}
+        coverage[fn]['parent_cov'] = percent_cov
+    
+    # Parse child coverage
+    root = ET.fromstring(child_data)
+    for elm in root.findall(".//File[@FullPath]"):
+        fn = elm.attrib["FullPath"]
+        if not want_file(fn):
+            continue
+        percent_cov = float(elm.find('PercentCoverage').text)
+        if fn not in coverage:
+            # After parsing parent, if the file stil do not exist, then the file is new
+            coverage[fn] = {'parent_cov': 0, 'new_file': True}
+        coverage[fn]['child_cov'] = percent_cov
+    
+    for fn in coverage:
+        coverage[fn]['delta_cov'] = coverage[fn]['child_cov'] - coverage[fn]['parent_cov']
+
+    return coverage    
+
+def parse_coverage_status(coverage):
+    """Return the exit code and exit message for given coverage
+    coverage: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
+    output: {"exit_code":int, "message":str} 
+    """
+    for fn in coverage:
+        if coverage[fn]["new_file"] and coverage[fn]["delta_cov"] <= 20:
+            # New files need to have at least 20% coverage
+            return {"exit_code":1, "message":"New files need at least 20% coverage."} 
+        elif coverage[fn]["delta_cov"] < 0 and abs(coverage[fn]["delta_cov"]) >= min(20, coverage[fn]["parent_cov"]):
+            return {"exit_code":1, "message":"Coverage decreased severely for more than 20% or a file lost coverage."} 
+        elif coverage[fn]["delta_cov"] < 0:
+            return {"exit_code":0, "message":"Coverage decreased slightly for less than 20%"} 
+    return {"exit_code":0, "message":"Coverage non decreasing."} 
+
+def coverage_table(coverage, format='txt'):
+    """Return the exit code and exit message for given coverage
+    coverage: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
+    output: None
+    """
+    columns = ["File Name", "Coverage (%)", "Delta Coverage"]
+    separator = ['-'*len(cn) for cn in columns]
+    result = [columns, separator] + [[fn, coverage[fn]["child_cov"], coverage[fn]["delta_cov"]] for fn in coverage]
+    return txt_table(result) if format == 'txt' else markdown_table(result)
+
+def main():
+    parser = argparse.ArgumentParser(description='Optional app description')
+    parser.add_argument('child_coverage', type=str,
+                        help='Child branch build directory')
+    parser.add_argument('parent_coverage', type=str,
+                        help='Parent branch build directory')
+    args = parser.parse_args()
+
+    child_cov_file = glob.glob(args.child_coverage + "/Testing/**/Coverage.xml", recursive = True)[0]
+    parent_cov_file = glob.glob(args.parent_coverage + "/Testing/**/Coverage.xml", recursive = True)[0]
+
+    with open(child_cov_file, 'r') as f:
+        child_data = f.read()
+    with open(parent_cov_file, 'r') as f:
+        parent_data = f.read()
+
+    aggregated_coverage = parse_coverage(child_data, parent_data)
+    txt_coverage_table = coverage_table(aggregated_coverage, format='txt')
+    exit_result = parse_coverage_status(aggregated_coverage)
+
+    print(txt_coverage_table)
+    print(exit_result["message"])
+    exit(exit_result["exit_code"])
+
+
+if __name__ == '__main__':
+    main()


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,104 @@
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+
+name: delta coverage checks
+
+on:
+  push:
```
pcolberg, line 9: This will only run once the pull request is merged into the `main` branch.

Unfortunately I could not find a better way than duplicating the paths for the `pull_request` event:
https://github.com/intel/fpga-runtime-for-opencl/blob/d78bdb41d2d84a42f511c14bb8fe42c1c061effe/.github/workflows/build-linux.yml#L17-L48

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,104 @@
...
+      TMP_DIR: /home/build/
+
+    steps:
+      - name: change ownership of workspace to current user
+        run: sudo chown -R build:build .
+
+      - name: checkout early-access
+        uses: actions/checkout@v2
+        with:
+          ref: early-access
```
pcolberg, line 45: The `ref` can be removed, to checkout the default branch (which is `main`).

.github/workflows/coverage.yml
sherry-yuan: > This will only run once the pull request is merged into the `main` branch.

You mean the docker pull will face permission issue until it is merged into main?

pcolberg: Good point. The pull request will run fine before being merged, but you need to include `package: read` permissions to pull the container image, which are private since these are only intended to be used in the CI workflows.
https://github.com/intel/fpga-runtime-for-opencl/blob/d78bdb41d2d84a42f511c14bb8fe42c1c061effe/.github/workflows/build-linux.yml#L8-L15

scripts/coverage_diff.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
```
pcolberg, line 1: All source files need to include a license header, for scripts you can use this format:
https://github.com/intel/fpga-runtime-for-opencl/blob/d78bdb41d2d84a42f511c14bb8fe42c1c061effe/scripts/install_aocl.sh#L1-L5
That is hash-bang, followed by license header, blank line, and then normal comments describing the file.

sherry-yuan:
Ready for review:
A failing check by deleting tests here: [Failing Check](https://github.com/intel/fpga-runtime-for-opencl/runs/4242001567?check_suite_focus=true) by deleting acl_device_test

Sorry I didn't notice when I click the 'make pull request' button it directly added you as reviewer. Should have marked it as draft earlier.

.github/workflows/coverage.yml
sherry-yuan: Thanks you! It worked! 

Sorry I didn't notice this from the other files

sherry-yuan:
Oops, I accidentally click the fetch upstream button from my repo, and somehow it does the merge. Let me try rebasing it instead.

.github/workflows/coverage.yml
pcolberg: No worries :slightly_smiling_face: I made the same mistake in the first version of https://github.com/intel/fpga-runtime-for-opencl/pull/16, **after** telling you about package permissions. :facepalm:

sherry-yuan:
Somehow after rebasing it shows the change that is already present in intel:main.....

pcolberg:
> Somehow after rebasing it shows the change that is already present in intel:main.....

The base of your pull request is still the older 248e37f2be0c62817ea6610172ac6ae16ba32fd2, not the newest d78bdb41d2d84a42f511c14bb8fe42c1c061effe.

I have these two remotes in my local repository:
```
git remote add intel https://github.com/intel/fpga-runtime-for-opencl
git remote add sherry-yuan https://github.com/sherry-yuan/fpga-runtime-for-opencl
```
Then update the remotes to the latest commit:
```
git fetch intel
git fetch sherry-yuan
```
Which yields this commit graph in `tig sherry-yuan/coverage intel/main`:
```
2021-11-16 13:49 -0800 Sherry Yuan   o [HEAD] {sherry-yuan/coverage} Add delta coverage check on pull requests
2021-11-16 14:35 -0500 Peter Colberg │ o {intel/main} {sherry-yuan/main} Support parsing JSON using jq in Linux build workflows
2021-11-15 19:09 -0500 Peter Colberg o─┘ Support Python in addition to Bash scripting in Linux build workflows
2021-11-15 15:17 -0500 Peter Colberg o Limit container image push to current tag
```
So you can update your pull-request branch using:
```
git rebase -i intel/main
```

sherry-yuan:
> So you can update your pull-request branch using:
> 
> ```
> git rebase -i intel/main
> ```

Thank you @pcolberg :) Yes I realized I just need to run it twice, I did the exact command the first time and it didn't work as expected.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,128 @@
...
+          CODE=$(cat "${{ env.TMP_DIR }}"/exit_code)
+          if [ \'"${CODE}"\' = \'2\' ]; then
+            echo "At least 1 file's coverage decrease by more than 20%"
+            exit 1
+          else
+            echo "No severe coverage decrease detected."
+          fi
+      - name: coverage non-decreasing
+        if: $(cat "${{ env.TMP_DIR }}"/exit_code) == '0'
+        run: echo "Coverage did not decrease."
```
pcolberg, line 128: On first look the resulting output of the exit code handling is nice. On the other hand it complicates the workflow quite a bit. Would you agree with simply invoking the python script, and letting the workflow fail without further output when the python scripts exits with a non-zero code? You print all the error cases within the script, which is perfectly sufficient.

The larger issue with workflows is that the workflow code cannot be tested outside of that environment, so I much prefer your python script over the additional workflow pretty printing. It's a balance, sometimes an inline [one-liner like this](https://github.com/intel/fpga-runtime-for-opencl/blob/a2b6ff4abf1385a477872cc85cc936521208ba5f/.github/workflows/unix.yml#L26-L33) is ok, but in this case the flow is more complicated and can live in the manually testable script.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Parse parent
+parent_file_to_coverage = {}
+root = ET.fromstring(parent_data)
+for elm in root.findall(".//File[@FullPath]"):
+    fn = elm.attrib["FullPath"]
+    if not want_file(fn):
+        continue
+    percent_cov = float(elm.find('PercentCoverage').text)
+    parent_file_to_coverage[fn] = percent_cov
```
pcolberg, line 57: This is a near-duplicate of the above block, which is a strong hint you can turn these into a common function :slightly_smiling_face:

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+import sys
+import argparse
+import glob
+
+parser = argparse.ArgumentParser(description='Optional app description')
+parser.add_argument('child_coverage', type=str,
+                    help='Child branch build directory')
+parser.add_argument('parent_coverage', type=str,
+                    help='Parent branch build directory')
+args = parser.parse_args()
```
pcolberg, line 19: You can move the argument parsing and other top-level code into a `main` function, which ensures the top-level variables are scoped to that function and cannot be used from other functions, which in turn ensures easier-to-comprehend functional code without global variables. Sometimes this is not necessary, but since you already have functions (also see duplication comment below) it does not hurt to go all the way with a `main` function.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+
+def want_file(fn):
+    return fn.startswith("./src")
+
+def pprint_table(matrix):
+    s = [[str(e) for e in row] for row in matrix]
+    lens = [max(map(len, col))+1 for col in zip(*s)]
+    fmt = '\t'.join('{{:{}}}'.format(x) for x in lens)
+    table = [fmt.format(*row) for row in s]
+    print('\n'.join(table))
```
pcolberg, line 32: Nitpick, but you could return a string here and leave the output method (here `print` to stdout) to the caller:
```python
print(pprint_table(matrix))
```

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+
+with open(child_cov_file, 'r') as f:
+    child_data = f.read()
+with open(parent_cov_file, 'r') as f:
+    parent_data = f.read()
```
pcolberg, line 37: In contact of below comment: These I would keep duplicate, since each is essentially a single function call.

Then your new function below would receive the `data` and returns the `coverage` dictionary.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+    child_cov = child_file_to_coverage[fn]
+    delta_cov = child_cov - parent_cov
+    result.append([fn, child_cov, delta_cov])
+    if fn not in parent_file_to_coverage and delta_cov <= 20:
+        # New files need to have at least 20% coverage
+        new_file_not_covered = True
+    if delta_cov < 0 and abs(delta_cov) >= min(20, parent_cov):
+        severe_decrease = True
+    if delta_cov < 0:
+        coverage_decreased = True
```
pcolberg, line 77: Please correct me if I am wrong here, but could this be broken up into roughly three steps?
1. A function to generate the `result` dict with key `fn` and values `parent_cov`, `child_cov`, and `child_cov`.
2. Print the `result` dict using `pprint_table`.
3. Iterate over the `result` dict, and if coverage is broken, print the status message and `exit(1)`.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+    print("New files are not covered")
+    exit(1)
+elif severe_decrease:
+    pprint_table(result)
+    print("Coverage decreased severely or a file lost coverage.")
+    exit(2)
+elif coverage_decreased:
+    pprint_table(result)
+    print("Coverage decreased.")
+    exit(78)
```
pcolberg, line 92: This special value would need an explanation, since one would naturally expect `3`.

But see below, exiting with `1` for any kind of failed coverage should be sufficient.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,96 @@
...
+else:
+    pprint_table(result)
+    print("Coverage did not decrease.")
+    exit(0)
```
pcolberg, line 96: As detailed above for the workflow, I would `exit(1)` when coverage fails (and do nothing when it passes).

.github/workflows/coverage.yml
pcolberg: Actually you want to remove the `push` event, and only trigger on `pull_request`. For the `push` event there is no base ref so nothing to compare to; we could generate a report for the main branch, but that is a separate workflow for later.

container/centos-8-dev/Dockerfile
```diff
@@ -21,6 +21,7 @@ RUN sed -i '/^enabled=/s#0#1#' /etc/yum.repos.d/CentOS-Linux-PowerTools.repo \
   gcc \
   gcc-c++ \
   git \
+  jq \
```
pcolberg, line 24: These should vanish once you rebase onto the latest `intel/main`.

pcolberg (changes requested):
Thanks @sherry-yuan, I am excited to integrate this into our pull-request checks.

The theme of this review is to functionalize all the things. :slightly_smiling_face:

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,128 @@
...
+    runs-on: ubuntu-20.04
+
+    container:
+      image: ghcr.io/intel/fpga-runtime-for-opencl/ubuntu-20.04-dev:main
+
+    continue-on-error: true
+
+    env:
+      WORK_DIR: /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl
+      TMP_DIR: /home/build/
```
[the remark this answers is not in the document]
pcolberg: Please never mind, using `$HOME` is not a good idea because the GitHub runner overrides with a directory, `/github/home` which is writable only to the root container user and differs from the home directory of the user.
- https://github.com/actions/runner/issues/863
- https://github.com/actions/runner/issues/1146

scripts/coverage_diff.py
sherry-yuan: The last few step of the workflow behave differently depending on the exit code of this python script, 78 exit code is used to be a "ignore status" for github workflow, although they now have removed it. 

The purpose of separating exit code is to let the workflow skip certain step. The workflow succeed but the step will look different than passing, just as a warning that the coverage has slightly decreased.

pcolberg: I see, thanks for the explanation of exit status 78. If that is no longer used, the only purpose of the different exit codes handling is to display different messages, correct? I still think a single workflow step, simply the running the script, would sufficient and more maintenable. Currently the script and workflow handle each case *three* times, where one time per case seems sufficient. Other than a pass or failed, there is no other result for a workflow, correct?

sherry-yuan: Right that make sense. I originally did it only for the "skip" symbol beside slight coverage decrease case. But I guess that doesn't bring too much value in terms of warning the developers. I have removed the extra workflow steps at the end.

pcolberg: > The workflow succeed but the step will look different than passing, just as a warning that the coverage has slightly decreased.

I missed this bit about the warning, now I understand your original intention. Unfortunately I would think that regardless of whether the warning is printed by the script, or shown as a separate workflow step, it would still be ignored. As long as the workflow shows a green check mark, I don't think anyone would check each time, unless they were specifically working on the workflow itself.

What do you think, would it be straight-forward to post a comment to the pull request from the python script with the coverage diff? That way the submitter has to acknowledge warnings, and also receives positive feedback when the coverage increases.

sherry-yuan: That's a great idea! Since PRs are a flavor of issue, the [issue comment API](https://docs.github.com/en/rest/reference/issues#create-an-issue-comment) will work

scripts/coverage_diff.py
sherry-yuan: Yes your understanding is correct, I have refractored things to be more modular.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,128 @@
...
+
+    steps:
+      - name: change ownership of workspace to current user
+        run: sudo chown -R build:build .
+
+      - name: checkout main branch
```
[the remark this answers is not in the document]
sherry-yuan: Base on the previous comments: "Actually you want to remove the push event, and only trigger on pull_request. For the push event there is no base ref so nothing to compare to" I have changed the workflow to trigger only on PR

scripts/coverage_diff.py
pcolberg: Great :slightly_smiling_face: For GitHub API usage, take a look at the different options and let me know which one(s) you prefer. A straight-forward option would be one of the python modules shipped with Ubuntu 20.04, e.g., [python3-github](https://packages.ubuntu.com/focal/python3-github) for high-level or [python3-requests](https://packages.ubuntu.com/focal/python3-requests) for low-level usage, but there might be better options.

.github/workflows/coverage.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         run: cd "$TMP_DIR/child_build" && mkdir -p Testing/random && touch Coverage.xml
 
       - name: coverage status
-        run: ./scripts/coverage_diff.py "$TMP_DIR/child_build" "$TMP_DIR/parent_build" --pr_id "${{ github.event.number }}"
+        run: ./scripts/coverage_diff.py "$TMP_DIR/child_build" "$TMP_DIR/parent_build" --pr_id "${{ github.event.number }}" --github_token "${{ secrets.GITHUB_TOKEN }}""
```
pcolberg: For better security, you can read the token from an environment variable in your script.
```yaml
run: ./scripts/coverage_diff.py "$TMP_DIR/child_build" "$TMP_DIR/parent_build" --pr_id '${{ github.event.number }}'
env:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
This way the token does not show up in the process list. This may not matter in practice if the runner is the only thing running on the system, but it is good practice regardless.

sherry-yuan:
Thank you for the suggestions, I came to realize that that the forked repo have maximum read [access](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) to anything through github token. The post comment functionality will only work when PRs are enabled directly in the public repo.

pcolberg:
> Thank you for the suggestions, I came to realize that that the forked repo have maximum read [access](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) to anything through github token. The post comment functionality will only work when PRs are enabled directly in the public repo.

Thanks @sherry-yuan for giving this a try, that is really unfortunate.

sherry-yuan:
An ugly workaround to post comment from forked to upstream is to upload comment in workflow artifact, and then have another cronjob workflow in the upstream repo to pick up artifacts and post comments.

https://github.com/nyurik/auto_pr_comments_from_forks

sherry-yuan:
I have changed everything back, ready for review.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,81 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    container:
+      image: ghcr.io/intel/fpga-runtime-for-opencl/ubuntu-20.04-dev:main
+
+    env:
+      TMP_DIR: /home/build/
```
pcolberg, line 41: Tiny nitpick: please remove trailing `/` which results in `//` when concatenated below. This is cosmetic only, but when `...//...` shows up in a log file, it may cause confusion as to whether something is wrong, e.g., use of an undefined variable `.../$MISSING/...`.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python3
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Install Intel FPGA SDK for OpenCL Pro Edition
+# https://fpgasoftware.intel.com/
```
pcolberg, line 6: Lines 5 and 6 can be removed, were copy-pasted over by accident from other script.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,144 @@
...
+
+# Example usage: python scripts/coverage_diff.py child_build_dir parent_build_dir
```
pcolberg, line 8: This would execute the script with a Python2 interpreter on many systems. How about
```python
 # Example usage: ./coverage_diff.py child_build_dir parent_build_dir
```
No need to duplicate the directory, script might be moved to another directory later then this is outdated.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,144 @@
...
+
+def markdown_table(array):
+    """ Input: Python list with rows of table as lists
+               First element as header. 
+        Output: String to put into a .md file 
+        
+    Ex Input: 
+        [["Name", "Age", "Height"],
+         ["Jake", 20, 5'10],
+         ["Mary", 21, 5'7]]
```
pcolberg, line 33: Are age and height also strings? `5'10` is not a valid Python expression.

As the example input, why not use (part of) the input data from an actual run.

scripts/coverage_diff.py
```diff
@@ -0,0 +1,144 @@
...
+               First element as header. 
+        Output: String to put into a .md file 
+        
+    Ex Input: 
+        [["Name", "Age", "Height"],
+         ["Jake", 20, 5'10],
+         ["Mary", 21, 5'7]] 
+    """
+
+    markdown = "\n" + str("| ")
```
pcolberg, line 36: You can use `"..."` instead of `str("...")` everywhere. `str()` is only needed when the input value is not a string.

pcolberg (changes requested):
Thanks @sherry-yuan, looks great :slightly_smiling_face: 

A few tiny nitpicks then it is ready to be merged.

pcolberg:
This is a suggestion for **after** this pull request has been merged: GitHub actions appears to support [source code annotations](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-warning-message) in workflows. I wonder if it works for forked pull requests too. Then you could annotate the coverage result to the top of each file for which coverage changed.

sherry-yuan:
> appears to support [source code annotations](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-warning-message) in workflows

Wow that's cool! Will do

All requested changes are applied.

pcolberg (approved):

PR #21, opened 2021-11-19
Check Delta Valgrind Logs

sherry-yuan:


sherry-yuan:
Caveat: valgrind suppression is too specific, sometimes it filter out memcheck of type Addr2 (invalid read of size 2), but Addr1,4,8 might also appear in the same place. 

Next step: add additional script for duplicating Addr memcheck types so that it won't miss the same valgrind issue.

sherry-yuan:
This is potentially a no go given the suppression are noisy and the valgrind errors are not consistent across the runs. Push to 2022.4 for now.

PR #22, opened 2021-11-19
Add source file annotation for coverage status

sherry-yuan:


sherry-yuan:
FYI annotation for unchanged file will not show up.

pcolberg:
> FYI annotation for unchanged file will not show up.

It shows here, maybe one needs to opt-into this beta feature somehow?
image

sherry-yuan:
> annotations from the python script itself

Phew, much easier than shell scripts.

sherry-yuan:
> opt into this beta feature somehow

I think it has both side to it, opting in for this feature might bring more noise, but on the other hand it bring more info in case any test files are changed.

sherry-yuan:
File annotation worked when adding uncovered lines in code! See this commit: https://github.com/intel/fpga-runtime-for-opencl/pull/22/commits/fb91da49f70c41d1340d29ce1c24daf105ca127b

sherry-yuan:
Will squash after review, in case the demo commit will vanish after squash.

scripts/coverage_diff.py
```diff
@@ -107,9 +107,16 @@ def coverage_table(coverage, format='txt'):
...
-    result = [columns, separator] + [[fn, coverage[fn]["child_cov"], coverage[fn]["delta_cov"]] for fn in coverage]
+    result = [columns, separator] + [[fn, coverage[fn]['child_cov'], coverage[fn]['delta_cov']] for fn in coverage]
     return txt_table(result) if format == 'txt' else markdown_table(result)
 
+def annotate_files(coverage):
+    """Return the exit code and exit message for given coverage
+    coverage: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
+    output: None
+    """
+    [print(f"::notice file={fn},line=1,endLine=1,title=Coverage::{coverage[fn]['child_cov']}% ({coverage[fn]['delta_cov']}%)") for fn in coverage if abs(coverage[fn]['delta_cov']) > 1]
```
pcolberg, line 118: Would it make sense to simply include the annotation in each `message` in `parse_coverage_status()`?

You could use different [message levels](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-notice-message) for each case, e.g.,
- `error` for ≥ 20% decrease
- `warning` for < 20% decrease
- `notice` for increase
- nothing if coverage is unchanged

pcolberg (changes requested):
Thanks @sherry-yuan. Along with the revision, please also add a temporary test commit to demonstrate the different cases. (I wonder if there is a better way than temporary commits to test workflows?)

sherry-yuan:
A commit demonstrating all three cases (severe decrease, slight decrease, increase) is in this commit https://github.com/intel/fpga-runtime-for-opencl/pull/22/commits/1cb17c2eb2d4e2b44fe646710226d2f64d8314ae

scripts/coverage_diff.py
sherry-yuan: I have separated out the exit status into a separate function so that it could be reused in post message. Just to keep things separated and organized.

sherry-yuan:
> add a temporary test commit to demonstrate the different cases.

I realized the annotation will not persist after new change are pushed. I will attach a printed pdf of the page.

pcolberg:
> I realized the annotation will not persist after new change are pushed. I will attach a printed pdf of the page.

That is fine, as long as your demo commit is in the history, I can push it when needed. Just delay the final squash until after the review to retain the demo commit for now.

sherry-yuan:
The coverage annotation can be viewed online here: https://www.scribd.com/document/541617197/Coverage-Annotation

scripts/coverage_diff.py
```diff
@@ -107,9 +105,28 @@ def coverage_table(coverage, format='txt'):
...
+    else:
+        return {"exit_code":0, "message":"Coverage non decreasing.", "log_level":"notice", "terminate":False}
+    
+def annotate_files(coverage):
+    """Return the exit code and exit message for given coverage
+    coverage: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
+    output: None
+    """
+    
+    [print(f"::{file_coverage_status(coverage[fn])['log_level']} file={fn},line=1,endLine=1,title=Coverage::{coverage[fn]['child_cov']}% ({coverage[fn]['delta_cov']}%)") for fn in coverage if abs(coverage[fn]['delta_cov']) > 1]
```
pcolberg, line 128: You can expand this loop to make it much easier to comprehend:
```python
for fn, cov in coverage.items():
  if not abs(cov['delta_cov']) > 1:
    continue

  status = file_coverage_status(cov)
  print(f"::{status['log_level']} file={fn},line=1,endLine=1,title=Coverage::{cov['child_cov']}% ({cov['delta_cov']}%)")
```
The idea is to write your code vertically, not horizontally.

scripts/coverage_diff.py
```diff
@@ -90,15 +90,13 @@ def parse_coverage_status(coverage):
     coverage: {filename: {parent_cov: float, child_cov: float, delta_cov: float, new_file: bool}}
     output: {"exit_code":int, "message":str} 
     """
+    latest_non_terminate_status = {"exit_code":0, "message":"Coverage non decreasing.", "log_level":"notice", "terminate":False}
```
pcolberg, line 93: This includes a `log_level`, but it is not used anywhere. That hints you something is not quite right with the organization of the functions, see next comment.

scripts/coverage_diff.py
```diff
@@ -90,15 +90,13 @@ def parse_coverage_status(coverage):
...
     for fn in coverage:
-        if coverage[fn]["new_file"] and coverage[fn]["delta_cov"] <= 20:
-            # New files need to have at least 20% coverage
-            return {"exit_code":1, "message":"New files need at least 20% coverage."} 
-        elif coverage[fn]["delta_cov"] < 0 and abs(coverage[fn]["delta_cov"]) >= min(20, coverage[fn]["parent_cov"]):
-            return {"exit_code":1, "message":"Coverage decreased severely for more than 20% or a file lost coverage."} 
-        elif coverage[fn]["delta_cov"] < 0:
-            return {"exit_code":0, "message":"Coverage decreased slightly for less than 20%"} 
-    return {"exit_code":0, "message":"Coverage non decreasing."} 
+        exit_status = file_coverage_status(coverage[fn])
```
pcolberg, line 95: Here you already have the `log_level`, so would it be possible to `print` the annotation right away?

scripts/coverage_diff.py
```diff
@@ -90,15 +90,13 @@ def parse_coverage_status(coverage):
...
+        if exit_status["terminate"]:
+            return exit_status
```
pcolberg, line 97: Then you won't be able to return early, so instead combine the statuses and return at the end of the function.

pcolberg (changes requested):
Looks good! Hopefully I did not miss anything obvious in the program organization.

scripts/coverage_diff.py
pcolberg: Would the following high-level structure work?
1. Convert coverage to coverage status by applying `file_coverage_status()` to all items in coverage.
2. Retrieve the combined exit status from coverage status.
3. Print the messages from coverage status.

sherry-yuan: I have combined things together, so now it will print all error message for each file separated by new line, and exit with the highest exit code returned from `file_coverage_status()`

sherry-yuan:
> That is fine, you can keep the demo commit in the history for the review
Thanks! Now the PR contain 2 commits, the first is the sqashed one ready for merge, and the last one is the demo change.

pcolberg:
Perfect, thanks @sherry-yuan. Would you be fine with removing `markdown_table`? The code is still in the history of the main branch for later reference, but it seems better not to have untested dead code, especially not in interpreted scripts.

pcolberg (approved):
Perfect, thanks @sherry-yuan, that's a nice solution :slightly_smiling_face:

sherry-yuan:
Thank you for the detailed review @pcolberg !

PR #31, opened 2021-12-03
Use default new C++11 ABI for std::string and std::list in libstdc++

pcolberg:
This flag was once used as a work-around for very old linkers and has
not been needed for a long time. The ABI change does not affect the
public runtime ABI since the runtime does not expose any C++ symbols.
The ABI change does affect the size of OpenCL data structures, however
these are not exposed directly but only in the form of opaque pointers.

https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html


zibaiwan (approved):
Looks good! Thanks @pcolberg .

pcolberg:
Thanks @zibaiwan. I have updated the commit message of the Klocwork fix after the test proved successful.

zibaiwan (approved):

PR #34, opened 2021-12-03
Add buffer location enum

sherry-yuan:
Sycl runtime calls clEnqueueWriteBuffer before clSetKernelArgs, this cause the buffer to be allocated on the device before knowing the right place of allocating the memory. As a result, later when kernel get invoked, the memory has to be copied from device's default global memory to the buffer location specified in kernel. This is an additional memory copy operation.

This extension does not interfer with the other way of setting buffer location (i.e through clSetKernelArgs). This property exist for integration with sycl runtime, not for pure opencl user to use. If opencl user wish to use this property, they have to make sure the buffer location passed into clCreateBufferWithPropertyINTEL has to match the one defined in kernel function interface, otherwise the extra memory copy issue will remain.

When resizing reserved allocation, we now have the information to allocate minimum amount of space required according to the property passed in.

### DIFF
--- a/src/acl_mem.cpp
+++ b/src/acl_mem.cpp
@@ -416,6 +416,7 @@ CL_API_ENTRY cl_mem clCreateBufferWithPropertiesINTEL(
   cl_bool context_has_device_with_physical_mem;
   unsigned int idevice;
   cl_uint bank_id = 0;
+  cl_uint tmp_mem_id = 0;
   acl_lock();
 
 #ifdef MEM_DEBUG_MSG
@@ -430,6 +431,9 @@ CL_API_ENTRY cl_mem clCreateBufferWithPropertiesINTEL(
                          "Both channel flag and channel property are set");
       }
       bank_id = (cl_uint) * (properties + 1);
+    } break;
+    case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+      tmp_mem_id = (cl_uint) * (properties + 1);
     } break;
     default: {
       UNLOCK_BAIL_INFO(CL_INVALID_DEVICE, context, "Invalid properties");
@@ -553,6 +557,7 @@ CL_API_ENTRY cl_mem clCreateBufferWithPropertiesINTEL(
     UNLOCK_BAIL_INFO(CL_OUT_OF_HOST_MEMORY, context,
                      "Could not allocate a cl_mem object");
   }
+  mem->mem_id = tmp_mem_id;
 
   mem->block_allocation = new_block;
   mem->block_allocation->mem_obj = mem;
@@ -784,7 +789,6 @@ CL_API_ENTRY cl_mem clCreateBufferWithPropertiesINTEL(
     mem->context = context;
     mem->flags = flags;
     mem->size = size;
-    mem->mem_id = 0;
 
     mem->bank_id = 0;
     if (is_SOC_device()) {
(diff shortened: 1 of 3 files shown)


sherry-yuan:
This change is almost ready to go, the enum made into [public khronos repo](https://github.com/KhronosGroup/OpenCL-Headers/blob/80c10b1f65b932894b830da7cd37bc56c541bae4/CL/cl_ext.h#L2350-L2361). I will make a separate PR to sync upstream headers, and then polish this a little.

sherry-yuan:
This change is ready to go.

test/acl_mem_test.cpp
```diff
@@ -2799,6 +2799,27 @@ TEST(acl_mem, case_205751_overlapping_alloc) {
...
+  size_t bank_size = total_size / 2;
+  size_t small_size = bank_size / 1024;
+
+  cl_mem_properties_intel props[] = {CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, 0, 0};
+  a = clCreateBufferWithPropertiesINTEL(m_context, props, 0, bank_size, 0,
+                                        &status);
+  CHECK_EQUAL(CL_SUCCESS, status);
+  ACL_LOCKED(CHECK_EQUAL(acl_bind_buffer_to_device(m_cq->device, a), 1));
+
+  CHECK_EQUAL(CL_SUCCESS, clReleaseMemObject(a));
```
pcolberg, line 2820: Would it be feasible to test `clCreateSubBuffer` here as well since it was modified above?

pcolberg (changes requested):
Thanks @sherry-yuan, great work!

Could you amend the commit message with links to the corresponding PRs in the OpenCL specification and header repositories?

src/acl_mem.cpp
```diff
@@ -416,6 +416,7 @@ CL_API_ENTRY cl_mem clCreateBufferWithPropertiesINTEL(
   cl_bool context_has_device_with_physical_mem;
   unsigned int idevice;
   cl_uint bank_id = 0;
+  unsigned int tmp_mem_id = 0;
```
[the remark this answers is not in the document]
sherry-yuan: Thanks for the suggestions, I applied the requested changes.

test/acl_mem_test.cpp
```diff
@@ -2799,6 +2799,57 @@ TEST(acl_mem, case_205751_overlapping_alloc) {
...
+      m_device[0]->def.autodiscovery_def.global_mem_defs[0].range);
+  size_t bank_size = total_size / 2;
+  size_t small_size = bank_size / 1024;
+
+  cl_mem_properties_intel props[] = {CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, 0, 0};
+  a = clCreateBufferWithPropertiesINTEL(m_context, props, 0, bank_size, 0,
+                                        &status);
+  ACL_LOCKED(CHECK(acl_mem_is_valid(a)));
+  CHECK_EQUAL(CL_SUCCESS, status);
+  if (a) {
```
pcolberg, line 2819: This would the test if for some reason `a` is `NULL`.

You can use `assert(a)` here and below to quiesce Klocwork.

pcolberg (changes requested):
Thanks Sherry for the testing! One minor change for Klocwork and it is ready to be merged.

test/acl_mem_test.cpp
sherry-yuan: Thank you! (sorry I totally forgot the convention about just using assert in test files). Now the changes are ready, and checks passed.

pcolberg (approved):

PR #37, opened 2021-12-06
Remove runtime assert for kernel finish_counter.

sherry-yuan:
Currently the kernel's finish_counter assert > 0 is called whenever there is a kernel interrupt sent. However the kernel interrupt could be sent whenever:
1. printf buffer is full
2. The kernel change state: init -> running -> finish/stall
3. clGetProfileInfoIntelFPGA is called

From the first reason, it already doesn't make sense to assert finish counter greater than 0, as is very possible for a printf buffer to be full before the kernel finishes.


pcolberg (approved):
Thanks @sherry-yuan!

PR #39, opened 2021-12-06
Revert "Use default new C++11 ABI for std::string and std::list in libstdc++"

pcolberg:
This change would require a larger ABI transition beyond the runtime.

This reverts commit 0605379d688d51eceb6204f6f0d7ae6a18293b28.


zibaiwan (approved):
Thank you @pcolberg

PR #43, opened 2021-12-14
Klocwork check null before dereference in acl_kernel_if

sherry-yuan:
Fix the following klocwork issues in acl_kernel_if.cpp:

1. Pointer 'kern->accel_csr' returned from call to function 'acl_malloc' at line 875 may be NULL and may be dereferenced at line 886.
2. Pointer 'kern->accel_perf_mon' returned from call to function 'acl_malloc' at line 877 may be NULL and may be dereferenced at line 898.
3. Pointer 'kern->accel_num_printfs' returned from call to function 'acl_malloc' at line 879 may be NULL and may be dereferenced at line 908.
4. Pointer 'kern->accel_job_ids' returned from call to function 'acl_malloc' at line 947 may be NULL and will be dereferenced at line 962.
5. Pointer 'kern->accel_queue_front' returned from call to function 'acl_malloc' at line 954 may be NULL and will be dereferenced at line 964.
6. Pointer 'kern->accel_invoc_queue_depth' returned from call to function 'acl_malloc' at line 949 may be NULL and will be dereferenced at line 961. 
7. Pointer 'kern->accel_queue_back' returned from call to function 'acl_malloc' at line 955 may be NULL and will be dereferenced at line 965.

### DIFF
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -949,6 +949,8 @@ int acl_kernel_if_update(const acl_device_def_autodiscovery_t &devdef,
   if (kern->num_accel > 0) {
     kern->accel_job_ids =
         (int volatile **)acl_malloc(kern->num_accel * sizeof(int *));
+    assert(kern->accel_job_ids);
+
     kern->accel_invoc_queue_depth =
         (unsigned int *)acl_malloc(kern->num_accel * sizeof(unsigned int));
 
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -874,6 +874,7 @@ int acl_kernel_if_update(const acl_device_def_autodiscovery_t &devdef,
     // Allocations for each kernel
     kern->accel_csr = (acl_kernel_if_addr_range *)acl_malloc(
         kern->num_accel * sizeof(acl_kernel_if_addr_range));
+    assert(kern->accel_csr);
     kern->accel_perf_mon = (acl_kernel_if_addr_range *)acl_malloc(
         kern->num_accel * sizeof(acl_kernel_if_addr_range));
     kern->accel_num_printfs =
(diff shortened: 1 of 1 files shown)


src/acl_kernel_if.cpp
```diff
@@ -874,10 +874,26 @@ int acl_kernel_if_update(const acl_device_def_autodiscovery_t &devdef,
...
+      ACL_KERNEL_IF_DEBUG_MSG(kern,
+                              "Failed to allocate memory for kernel csr.\n");
+      return 1;
+    }
     kern->accel_perf_mon = (acl_kernel_if_addr_range *)acl_malloc(
         kern->num_accel * sizeof(acl_kernel_if_addr_range));
+    if (!(kern->accel_perf_mon)) {
+      ACL_KERNEL_IF_DEBUG_MSG(
+          kern, "Failed to allocate memory for kernel performance monitor.\n");
+      return 1;
```
pcolberg, line 887: The `return` leaks memory allocated previously, e.g., for `kern->accel_csr`.

There are two options here:
1. You can either clean up the resources meticulously before each return. In C programming, this is usually implemented by having a resource clean-up block at the end of the function, and jumping to the appropriate line within that block using `goto`. In C++ programming, this is resolved using [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization), where the resource is managed by an object with a deconstructor freeing the resource.

2. Instead, I would suggest to simply abort with `assert()`. In an out-of-memory situation, it is usually difficult to recover gracefully anyway, so aborting does not make things any worse.

pcolberg (changes requested):
Thanks @sherry-yuan! I would suggest to simply go with `assert()`.

src/acl_kernel_if.cpp
sherry-yuan: Thanks, I have changed all if statements to asserts instead.

PR #45, opened 2021-12-15
Show fixed Klocwork Issue Without Rerunning Analysis

sherry-yuan:
The change can be described as follows:
1. It first gets the detailed issue from parent build
2. get detailed issue from child build
3. extract the issue id from both build see what is unique to just the child build (new klocwork issues) and what is unique to only the parent build (fixed klocwork issues) | compare the ids instead of issue description because file lines may change
4. Return the blocks that contain the pattern `(id1|id2|id3) \(Local\)`, each block is separated by long '----------------'

### DIFF
--- a/.github/workflows/klocwork-pull-request.yml
+++ b/.github/workflows/klocwork-pull-request.yml
@@ -46,7 +46,9 @@ jobs:
         run: sudo chown -R klocwork:klocwork .
 
       - name: create local Klocwork project
-        run: kwcheck create --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+        run: |
+          mkdir -p /home/klocwork/parent
+          kwcheck create --project-dir /home/klocwork/parent/.kwlp --settings-dir /home/klocwork/parent/.kwps
 
       - name: checkout base ref
         uses: actions/checkout@v2
@@ -62,12 +64,16 @@ jobs:
         run: kwinject --output kwinject.out ninja -C build -v -k0
 
       - name: fully analyze build of base ref
-        run: kwcheck run --build-spec kwinject.out -F scriptable --report issues.txt --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+        run: kwcheck run --build-spec kwinject.out -F scriptable --report /home/klocwork/parent_issues.txt --project-dir /home/klocwork/parent/.kwlp --settings-dir /home/klocwork/parent/.kwps
 
-      - name: ignore issues of base ref
+      # Use non numeric sort because `comm` expects default sorting order
+      - name: get parent issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/parent_issues.txt | sort > /home/klocwork/parent_issues_ids.txt
+      
+      - name: copy parent klocwork project
         run: |
-          count=$(wc -l < issues.txt)
-          test "$count" = "0" || kwcheck set-status 1-"$count" --status ignore --project-dir /home/klocwork/.kwlp
+          mkdir -p /home/klocwork/child
+          cp -r /home/klocwork/parent/.kwlp /home/klocwork/parent/.kwps /home/klocwork/child
 
       - name: checkout head ref
         uses: actions/checkout@v2
@@ -81,13 +87,31 @@ jobs:
         run: kwinject --output kwinject.out ninja -C build -v -k0
 
       - name: incrementally analyze build of head ref
-        run: kwcheck run --build-spec kwinject.out -F scriptable --report issues.txt --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+        run: kwcheck run --build-spec kwinject.out -F scriptable --report /home/klocwork/child_issues.txt --project-dir /home/klocwork/child/.kwlp --settings-dir /home/klocwork/child/.kwps
+
+      # Klocwork incremental scans will suppress the issues that were fixed in head ref (while preserving their issue id)
+      # Any new issue will have id will be larger than the max id in scan of base ref.
+      # For example, if base ref scan produced issue ids 1-5, and in head ref issue 4 was resolved but also introduced a new issue
+      # then the resulting issue ids for head ref scan will be 1,2,3,5,6.
+      - name: get child issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
+      
+      - name: list issues fixed in head ref
+        run: |
+          FIXED_ISSUE_ID=$(comm -13 /home/klocwork/child_issues_ids.txt /home/klocwork/parent_issues_ids.txt | paste -sd,)
+          if [ -n "$FIXED_ISSUE_ID" ]
+          then
+            kwcheck list --id "$FIXED_ISSUE_ID" -F detailed --project-dir /home/klocwork/parent/.kwlp --settings-dir /home/klocwork/parent/.kwps
+          fi
 
       - name: ensure no new issues have been found in head ref
         run: |
-          kwcheck list -F detailed --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
-          size=$(stat --printf=%s issues.txt)
-          test "$size" = "0"
+          NEW_ISSUE_ID=$(comm -13 /home/klocwork/parent_issues_ids.txt /home/klocwork/child_issues_ids.txt | paste -sd,)
+          if [ -n "$NEW_ISSUE_ID" ]
+          then
+            kwcheck list --id "$NEW_ISSUE_ID" -F detailed --project-dir /home/klocwork/child/.kwlp --settings-dir /home/klocwork/child/.kwps
+          fi
+          test -z "$NEW_ISSUE_ID"
 
       - name: revert ownership of workspace to root
         run: sudo chown -R root:root .


.github/workflows/klocwork-pull-request.yml
```diff
@@ -102,17 +102,24 @@ jobs:
...
       - name: get child issue ids
-        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt |sort > /home/klocwork/child_issues_ids.txt
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
       
       - name: fixed issues
         run: |
           FIXED_ISSUE_ID=$(comm -13 /home/klocwork/child_issues_ids.txt /home/klocwork/parent_issues_ids.txt | sed -z 's/\n/,/g;s/,$/\n/')
-          kwcheck list --id ${FIXED_ISSUE_ID} -F detailed --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+          echo kwcheck list --id ${FIXED_ISSUE_ID} -F detailed --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+          if [ \"${FIXED_ISSUE_ID}\" != \"\" ]
```
pcolberg, line 110: The escaping `\` is not needed. In bash ([which is the default](https://github.com/actions/runner/blob/main/docs/adrs/0277-run-action-shell-options.md)) you can use `-v` to check if a variable is set:
```
if [[ -v FIXED_ISSUE_ID ]]; then
```
(This has the advantage that it also works in `set -u` mode, though that is not the default in GitHub actions.)

.github/workflows/klocwork-pull-request.yml
```diff
@@ -102,17 +102,24 @@ jobs:
           if-no-files-found: error
       
       - name: get child issue ids
-        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt |sort > /home/klocwork/child_issues_ids.txt
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
       
       - name: fixed issues
         run: |
           FIXED_ISSUE_ID=$(comm -13 /home/klocwork/child_issues_ids.txt /home/klocwork/parent_issues_ids.txt | sed -z 's/\n/,/g;s/,$/\n/')
```
pcolberg, line 109: Not entirely sure if this applies here, but there is a nice shortcut to concatenate lines in this manner:
```
paste -sd, <<'EOF'
a
b
c
d
EOF
```
```
a,b,c,d
```

sherry-yuan: Thanks! Its much readable and simpler.

.github/workflows/klocwork-pull-request.yml
sherry-yuan: In this case I think the variable is set regardless, I have now used -n option for checking whether the given env var is empty.

src/acl_mem.cpp
```diff
@@ -2031,6 +2031,11 @@ CL_API_ENTRY cl_int CL_API_CALL clEnqueueReadImageIntelFPGA(
                    "Pointer argument cannot be NULL");
   }
 
+  if (!image || !is_image(image)) {
+    UNLOCK_ERR_RET(CL_INVALID_MEM_OBJECT, command_queue->context,
+                   "Memory object is not an image");
+  }
+
```
sherry-yuan, line 2038: Fixed a klocwork issue, the output is 

```
-----
160 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/src/acl_mem.cpp:2058 NPD.CHECK.MUST (1:Critical) Analyze
Pointer 'image' checked for NULL at line 2034 will be dereferenced at line 2058.
  * acl_mem.cpp:2034: 'image' is checked for NULL.
  * acl_mem.cpp:2034: image!=0 is false
  * acl_mem.cpp:2058: 'image' is explicitly dereferenced.
Current status 'Analyze'


-----
161 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/src/acl_mem.cpp:2065 NPD.CHECK.MUST (1:Critical) Analyze
Pointer 'image' checked for NULL at line 2034 will be dereferenced at line 2065.
  * acl_mem.cpp:2034: 'image' is checked for NULL.
  * acl_mem.cpp:2034: image!=0 is false
  * acl_mem.cpp:2065: 'image' is explicitly dereferenced.
Current status 'Analyze'
```

test/acl_profiler_test.cpp
```diff
@@ -49,6 +49,10 @@ MT_TEST_GROUP(acl_profile) {
   // Preloads a standard platform, and finds all devices.
   enum { MAX_DEVICES = 100, m_num_devices_in_context = 3 };
   void setup() {
+    int *a;
+    *a = 1;
+    int *b = NULL;
+    *b = 2;
```
sherry-yuan, line 55: Add intentional klocwork issues, the output is

```
-----
174 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/test/acl_profiler_test.cpp:55 NPD.GEN.MUST (1:Critical) Analyze
Null pointer 'b' that comes from line 54 will be dereferenced at line 55.
  * acl_profiler_test.cpp:54: 'b' has been assigned a NULL value.
  * acl_profiler_test.cpp:55: 'b' is explicitly dereferenced.
Current status 'Analyze'


-----
175 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/test/acl_profiler_test.cpp:53 UNINIT.STACK.MUST (1:Critical) Analyze
'a' is used uninitialized in this function.
  * acl_profiler_test.cpp:52: 'a' is declared
  * acl_profiler_test.cpp:53: 'a' is used, but is uninitialized.
Current status 'Analyze'
```

sherry-yuan:
When nothing is changed, both fixed issue, and new issue step should give no output.

sherry-yuan:
The change can be described as follows:
1. It first gets the detailed issue from parent build
2. get detailed issue from child build
3. extract the issue id from both build see what is unique to just the child build (new klocwork issues) and what is unique to only the parent build (fixed klocwork issues) | compare the ids instead of issue description because file lines may change
4. Use awk to pattern match `(id1|id2|id3) \(Local\)` and return back block separated by long '----------------'

.github/workflows/klocwork-pull-request.yml
```diff
@@ -81,13 +89,37 @@ jobs:
...
 
       - name: ensure no new issues have been found in head ref
         run: |
-          kwcheck list -F detailed --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
-          size=$(stat --printf=%s issues.txt)
-          test "$size" = "0"
+          NEW_ISSUE_ID=$(comm -13 /home/klocwork/parent_issues_ids.txt /home/klocwork/child_issues_ids.txt | paste -sd'|')
+          if [ -n "${NEW_ISSUE_ID}" ]
+          then
+            awk -vRS='----------' "/(${NEW_ISSUE_ID}) \(Local\)/{print}" /home/klocwork/child_issues_details.txt
```
pcolberg, line 120: Why are the input records separated by
```
---------------------------------------------------------------------------
```
whereas the output records are separated by
```
-----
```
and the record separator is `----------`?

`(^|\n)-{75}\n` should match the record separator exactly, e.g.:
```
awk -vRS='(^|\n)-{75}\n' "/^(1|23) \(Local\)/{print}" child_issues_details.txt
```
```
1 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/test/acl_svm_test.cpp:382 NPD.FUNC.MUST (1:Critical) Analyze
Pointer 'src_ptr' returned from call to function 'clSVMAlloc' at line 371 may be NULL and will be dereferenced at line 382.
  * acl_svm_test.cpp:372: 'src_ptr' is assigned the return value from function 'clSVMAlloc'.
    * acl_svm.cpp:171: 'clSVMAllocIntelFPGA(context, flags, size, alignment)' is assigned the return value from function 'clSVMAllocIntelFPGA'.
      * acl_svm.cpp:51: 'clSVMAllocIntelFPGA' explicitly returns a NULL value.
    * acl_svm.cpp:171: 'clSVMAllocIntelFPGA(context, flags, size, alignment)' is returned by 'clSVMAlloc'.
  * acl_svm_test.cpp:382: 'src_ptr' is explicitly dereferenced.
Current status 'Analyze'

23 (Local) /__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl/lib/pkg_editor/src/pkg_editor.c:278 NPD.FUNC.MIGHT (1:Critical) Analyze
Pointer 'data' returned from call to function 'get_name_data_ptr' at line 269 may be NULL and may be dereferenced at line 278.
  * pkg_editor.c:269: 'data' is assigned the return value from function 'get_name_data_ptr'.
    * pkg_editor.c:133: 'get_name_data_ptr' explicitly returns a NULL value.
  * pkg_editor.c:278: 'data' is explicitly dereferenced.
Current status 'Analyze'
```

pcolberg: I like your creative use of `awk` :shell: :ninja:, but I am unsure whether it is a good idea to parse the `detailed` output at all, since that is what `scriptable` (or `json`) output is intended for. [`kwcheck list`](https://docs.roguewave.com/en/klocwork/current/kwcheck#kwchecklist) has an `--id` option that allows specifying a comma-separated list of issues. Would that satisfy your goal to list a subset of the issues?

pcolberg (changes requested):
Thanks @sherry-yuan, this will be very useful when fixing existing Klocwork issues!

.github/workflows/klocwork-pull-request.yml
```diff
@@ -62,12 +62,20 @@ jobs:
...
 
-      - name: ignore issues of base ref
-        run: |
-          count=$(wc -l < issues.txt)
-          test "$count" = "0" || kwcheck set-status 1-"$count" --status ignore --project-dir /home/klocwork/.kwlp
+      - name: get parent issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/parent_issues.txt | sort > /home/klocwork/parent_issues_ids.txt
+
+      - name: get detailed parent issue
+        run: kwcheck list -F detailed --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps | head -n -3 > /home/klocwork/parent_issues_details.txt
```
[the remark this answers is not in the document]
pcolberg: `head -n -3` is probably no longer needed if you use `--id` (below), if the `detailed` output does not need to be parsed.

.github/workflows/klocwork-pull-request.yml
```diff
@@ -81,13 +89,37 @@ jobs:
...
+      
+      - name: get child issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
+      
+      - name: fixed issues
+        run: |
+          FIXED_ISSUE_ID=$(comm -13 /home/klocwork/child_issues_ids.txt /home/klocwork/parent_issues_ids.txt | paste -sd'|')
+          if [ -n "${FIXED_ISSUE_ID}" ]
+          then
+            awk -vRS='----------' "/(${FIXED_ISSUE_ID}) \(Local\)/{print}" /home/klocwork/parent_issues_details.txt
```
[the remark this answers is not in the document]
pcolberg: `^` matches the beginning of the record, so that only works if `RS` consumes the record separator exactly.

.github/workflows/klocwork-pull-request.yml
sherry-yuan: The caveat with using kwcheck, which I realized when I was testing the flow is: When the issue is resolved in the child build, the issue will be suppressed (although its issue id is preserved). For example there were original issue 1-5, and issue 4 was resolved, but a new issue was also discovered, then the resulting child issues will be 1,2,3,5,6. This means if we try to list the fixed issues (4), then this will not show.

sherry-yuan: > awk -vRS='(^|\n)-{75}\n' "/^(1|23) \(Local\)/{print}" child_issues_details.txt

Thank you, yes this works.

.github/workflows/klocwork-pull-request.yml
```diff
@@ -81,13 +89,37 @@ jobs:
...
+      - name: upload klockwork issues
+        uses: actions/upload-artifact@v2
+        with:
+          name: klocwork-${{ matrix.image }}-${{ matrix.config }}-${{ github.run_id }}
+          path: /home/klocwork/child_issues_details.txt
+          if-no-files-found: error
+      
+      - name: get child issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
+
```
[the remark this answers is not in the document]
sherry-yuan: Thank you for reviewing:) I have added the necessary documentations above some of the steps.

.github/workflows/klocwork-pull-request.yml
pcolberg: Thanks @sherry-yuan, I verified that even
```
kwcheck set-status --status Analyze 123
kwcheck list --id 123
```
does not list a resolved issue.

In that case, what do you think of having separate paths for the parent and child Klocwork projects, and copying the parent to the child project before the child analysis? The current solution fails silently when the `detailed` output format changes in a new Klocwork version, which should be expected since it is not intended for parsing.

.github/workflows/klocwork-pull-request.yml
```diff
@@ -62,12 +62,20 @@ jobs:
...
       - name: fully analyze build of base ref
-        run: kwcheck run --build-spec kwinject.out -F scriptable --report issues.txt --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
+        run: kwcheck run --build-spec kwinject.out -F scriptable --report /home/klocwork/parent_issues.txt --project-dir /home/klocwork/.kwlp --settings-dir /home/klocwork/.kwps
 
-      - name: ignore issues of base ref
-        run: |
-          count=$(wc -l < issues.txt)
-          test "$count" = "0" || kwcheck set-status 1-"$count" --status ignore --project-dir /home/klocwork/.kwlp
+      - name: get parent issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/parent_issues.txt | sort > /home/klocwork/parent_issues_ids.txt
```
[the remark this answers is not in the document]
pcolberg: Actually your initial solution using non-numeric `sort` was better, since that way you can make use of the (default) ordering check in `comm`. So I would suggest you revert to non-numeric `sort`, but add a comment **why** the sort is non-numeric.

.github/workflows/klocwork-pull-request.yml
sherry-yuan: Thank you! This is a great idea and it worked 🎉

.github/workflows/klocwork-pull-request.yml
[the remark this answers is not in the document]
sherry-yuan: I have added the comment and changed back to default sort.

.github/workflows/klocwork-pull-request.yml
```diff
@@ -62,12 +65,16 @@ jobs:
...
+      # Use non numeric sort because comm expect default sorting order
+      - name: get parent issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/parent_issues.txt | sort > /home/klocwork/parent_issues_ids.txt
+      
+      - name: copy parent klocwork project
         run: |
-          count=$(wc -l < issues.txt)
-          test "$count" = "0" || kwcheck set-status 1-"$count" --status ignore --project-dir /home/klocwork/.kwlp
+          cp -r /home/klocwork/parent/.kwlp /home/klocwork/child/.kwlp 
+          cp -r /home/klocwork/parent/.kwps /home/klocwork/child/.kwps
```
pcolberg, line 77: This may be combined:
```
cp -r /home/klocwork/parent/.kwlp /home/klocwork/parent/.kwps /home/klocwork/child
```
(Bash would also allow for
```
cp -r /home/klocwork/parent/.kw{lp,ps} /home/klocwork/child
```
which is nice for interactive use, but in scripts I would stick to the first, portable form.)

.github/workflows/klocwork-pull-request.yml
```diff
@@ -81,13 +88,50 @@ jobs:
...
+          if-no-files-found: error
+
+      # Klocwork incremental scans will suppress the issues that were fixed in head ref (while preserving their issue id)
+      # Any new issue will have id will be larger than the max id in scan of base ref.
+      # For example, if base ref scan produced issue ids 1-5, and in head ref issue 4 was resolved but also introduced a new issue
+      # then the resulting issue ids for head ref scan will be 1,2,3,5,6.
+      - name: get child issue ids
+        run: awk -F ';' '{print $1}' /home/klocwork/child_issues.txt | sort > /home/klocwork/child_issues_ids.txt
+      
+      - name: fixed issues
```
pcolberg, line 119: How about `list issues fixed in head ref`?

pcolberg (changes requested):
Perfect, thanks @sherry-yuan!

pcolberg (approved):

PR #46, opened 2021-12-16
Buffer location property for USM device allocation

sherry-yuan:
This is currently rebased upon buffer_location branch.

Related PR: #34 (34 has to be merged in first)

### DIFF
--- a/src/acl_usm.cpp
+++ b/src/acl_usm.cpp
@@ -239,10 +239,14 @@ clDeviceMemAllocINTEL(cl_context context, cl_device_id device,
   // Iterate over properties.
   // The end of the properties list is specified with a zero.
   cl_mem_alloc_flags_intel alloc_flags = 0;
+  cl_uint mem_id = acl_get_default_memory(device->def);
   while (properties != NULL && *properties != 0) {
     switch (*properties) {
     case CL_MEM_ALLOC_FLAGS_INTEL: {
       alloc_flags = *(properties + 1);
+    } break;
+    case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+      mem_id = (cl_uint) * (properties + 1);
     } break;
     default: {
       UNLOCK_BAIL_INFO(CL_INVALID_DEVICE, context, "Invalid properties");
@@ -254,8 +258,10 @@ clDeviceMemAllocINTEL(cl_context context, cl_device_id device,
   cl_int status;
 
   // Use cl_mem for convenience
+  cl_mem_properties_intel props[] = {CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, mem_id,
+                                     0};
   cl_mem usm_device_buffer = clCreateBufferWithPropertiesINTEL(
-      context, NULL, CL_MEM_READ_WRITE, size, NULL, &status);
+      context, props, CL_MEM_READ_WRITE, size, NULL, &status);
   if (status != CL_SUCCESS) {
     UNLOCK_BAIL_INFO(status, context, "Failed to allocate device memory");
   }
@@ -602,6 +608,14 @@ CL_API_ENTRY cl_int CL_API_CALL clGetMemAllocInfoINTEL(
       RESULT_UINT(usm_alloc->type);
     } else {
       RESULT_UINT(CL_MEM_TYPE_UNKNOWN_INTEL);
+    }
+  } break;
+
+  case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+    if (usm_alloc) {
+      RESULT_UINT(usm_alloc->mem->mem_id);
+    } else {
+      RESULT_UINT(0);
     }
   } break;
 
--- a/test/acl_usm_test.cpp
+++ b/test/acl_usm_test.cpp
@@ -388,6 +388,62 @@ MT_TEST(acl_usm, alloc_and_free_device_usm) {
   ACL_LOCKED(acl_print_debug_msg("end alloc_and_free_device_usm\n"));
 }
 
+MT_TEST(acl_usm, buffer_location_usm) {
+  ACL_LOCKED(acl_print_debug_msg("begin buffer_location_usm\n"));
+  const int alignment = ACL_MEM_ALIGN;
+  cl_int status;
+  this->yeah = true;
+
+  acl_usm_allocation_t *test_device_alloc;
+
+  cl_mem_properties_intel good_property[3] = {
+      CL_MEM_ALLOC_BUFFER_LOCATION_INTEL, 0, 0};
+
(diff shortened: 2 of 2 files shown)


src/acl_usm.cpp
```diff
@@ -606,6 +612,14 @@ CL_API_ENTRY cl_int CL_API_CALL clGetMemAllocInfoINTEL(
     }
   } break;
 
+  case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+    if (usm_alloc) {
+      RESULT_UINT(usm_alloc->mem->mem_id);
+    } else {
+      RESULT_UINT(CL_MEM_TYPE_UNKNOWN_INTEL);
```
pcolberg, line 619: Where is this return value specified? The [cl_intel_mem_alloc_buffer_location specification](https://github.com/KhronosGroup/OpenCL-Docs/blob/591eb712d8d9a7d47d1d48437ae84521a3c03121/extensions/cl_intel_mem_alloc_buffer_location.asciidoc#modifications-to-the-opencl-api-specification) mentions this:
> Returns buffer location for the Unified Shared Memory allocation.
>
> Return the ID of default global memory if no buffer location property were specified.

test/acl_usm_test.cpp
```diff
@@ -388,6 +388,52 @@ MT_TEST(acl_usm, alloc_and_free_device_usm) {
...
+  ACL_LOCKED(CHECK_EQUAL(acl_usm_ptr_belongs_to_context(m_context, test_ptr),
+                         CL_TRUE));
+  ACL_LOCKED(test_device_alloc =
+                 acl_get_usm_alloc_from_ptr(m_context, test_ptr));
+  assert(test_device_alloc);
+  CHECK_EQUAL(test_device_alloc->range.begin, test_ptr);
+
+  // Check alloc information
+  cl_uint read_mem_id = 4;
+  size_t ret_size;
```
pcolberg, line 420: This variable should be initialized and checked too.

test/acl_usm_test.cpp
```diff
@@ -388,6 +388,52 @@ MT_TEST(acl_usm, alloc_and_free_device_usm) {
...
+                                  sizeof(cl_uint), &read_mem_id, &ret_size);
+
+  CHECK_EQUAL(CL_SUCCESS, status);
+  CHECK_EQUAL(0, read_mem_id);
+
+  status = clMemFreeINTEL(m_context, test_ptr);
+  CHECK_EQUAL(status, CL_SUCCESS);
+  ACL_LOCKED(CHECK_EQUAL(acl_usm_ptr_belongs_to_context(m_context, test_ptr),
+                         CL_FALSE));
+  CHECK(m_context->usm_allocation.empty());
```
pcolberg, line 432: Would it make sense to check `clGetMemAllocInfoINTEL` once more, for the case where `usm_alloc` is `NULL`?

pcolberg (changes requested):
Thanks @sherry-yuan for the thorough testing :slightly_smiling_face:

src/acl_usm.cpp
sherry-yuan: I think the spec is meant to say when no property is passed into clDeviceMemAlloc, then default mem is returned, but in this case if the usm pointer itself is null, the behavior is not defined in the spec. I will just make it return 0 (I am surprised that this function does not error out when usm pointer is null).

test/acl_usm_test.cpp
sherry-yuan: Test case added.

test/acl_usm_test.cpp
```diff
@@ -388,6 +388,61 @@ MT_TEST(acl_usm, alloc_and_free_device_usm) {
...
+  size_t ret_size = 9;
+  status = clGetMemAllocInfoINTEL(m_context, test_ptr,
+                                  CL_MEM_ALLOC_BUFFER_LOCATION_INTEL,
+                                  sizeof(cl_uint), &read_mem_id, &ret_size);
+
+  CHECK_EQUAL(CL_SUCCESS, status);
+  CHECK_EQUAL(0, read_mem_id);
+  CHECK_EQUAL(sizeof(cl_uint), ret_size);
+  // Check when given pointer is invalid
+  status = clGetMemAllocInfoINTEL(m_context, NULL,
```
pcolberg, line 429: How about testing `test_ptr` after you free it below?

src/acl_usm.cpp
```diff
@@ -606,6 +612,14 @@ CL_API_ENTRY cl_int CL_API_CALL clGetMemAllocInfoINTEL(
     }
   } break;
 
+  case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+    if (usm_alloc) {
+      RESULT_UINT(usm_alloc->mem->mem_id);
+    } else {
+      RESULT_UINT(0);
```
pcolberg, line 619: Is `0` the "ID corresponding to default global memory"? If so, this would be a suitable comment here.

pcolberg (changes requested):

src/acl_usm.cpp
sherry-yuan: Technically there isn't an buffer_location info of a null or invalid usm. But for any valid usm that do not have property specified, it is gotten from `acl_get_default_memory(device->def)` Which is the memory in board_spec.xml annotated with default.

So default memory is not necessarily 0, its just that when usm is null, there is no information about it, so I just return 0.

pcolberg: Thanks @sherry-yuan. Could you clarify with the `buffer_location` specification contributors what the desired behaviour would be when the pointer does not point to a valid usm allocation? Should an error be returned instead? Does the specification need clarification?

sherry-yuan: I think it is fine to return 0, because global mem 0 always exist in the board spec, and for user who expect non zero answer, they probably already know something is wrong with their pointer when it is not. Further it is almost a convention that developers checks for validity of the usm pointers before running any other function on it.

I don't think there need to be a spec change, as 0 is the assumed default mem (when no other mem are annotated).

sherry-yuan:
I think something is going wrong with my permission of accessing klocwork. (see klocwork scans)

pcolberg:
> I think something is going wrong with my permission of accessing klocwork. (see klocwork scans)

No worries, the runner machine had been rebooted and the network connection temporarily failed when the workflow run launched. I restarted the workflow and it seems fine now.

src/acl_usm.cpp
sherry-yuan: Another reason is this spec is general to all vendors. Other vendor's default mem may have other values, which then it doesn't make sense to forcefully set it 0 in this case.

pcolberg (approved):

PR #49, opened 2021-12-16
Link acl target with pthread library

pcolberg:
This resolves a link failure with `-Wl,--no-undefined`.


zibaiwan (approved):
Thanks @pcolberg

PR #52, opened 2022-01-05
Readability: Comment struct field beside struct initialization

sherry-yuan:
--------------------------------------------------------------------

Previously when we navigate the code, we usually see calls like `acl_get_hal()->shared_alloc`, however when we search for `shared_alloc` we do not see where it is defined immediately. This has caused many confusion when navigating the code.

By adding the struct field corresponding to each definition, it help us to directly find what function corresponds to it without matching the index of the struct.

### DIFF
--- a/src/acl_hal_mmd.cpp
+++ b/src/acl_hal_mmd.cpp
@@ -302,47 +302,49 @@ acl_kernel_update_callback acl_kernel_update_fn = NULL;
 acl_profile_callback acl_profile_fn = NULL;
 acl_device_update_callback acl_device_update_fn = NULL;
 
-static acl_hal_t acl_hal_mmd = {acl_hal_mmd_init_device,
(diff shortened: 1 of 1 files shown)


sherry-yuan:
Realized the struct cannot use designated initialization because windows requires c++ 20+ to do this.

zibaiwan (approved):
Thanks @sherry-yuan for increasing the readability. Your change looks good to me. In the beginning, I was thinking maybe each individual comment can be replaced by a top comment like  "all acl_hal_mmd_<function name> is mapped to the <function name> in the struct". After a second thought, I think your way makes the search easier, when people try to search for "function name" only.

sherry-yuan:
Thank you for the review @zibaiwan ! Is it ok if you merge this in? I think I do not have the access to (I don't see merge button from my end).

PR #53, opened 2022-01-07
Implement OpenCL Core 3.0 clCreateBufferWithProperties

### LINKED ISSUE #51: Implement clCreateBufferWithProperties for OpenCL Core 3.0
In the most recently imported headers, there is a new function called `clCreateBufferWithProperties` which is equivalent to our `clCreateBufferWithPropertiesINTEL` extension function.
https://github.com/intel/fpga-runtime-for-opencl/blob/950f21dd079dfd55a473ba4122a4a9dca450e36f/include/CL/cl.h#L1124

We need to support/implement this function.

sherry-yuan:
In the most recently imported headers, there is a new function called clCreateBufferWithProperties which is equivalent to our clCreateBufferWithPropertiesINTEL extension function. Implement this function.

Closes #51


sherry-yuan:
Let me know if I should wait until all implementation of core 3.0 is done before making this ready for review.

src/acl_mem.cpp
```diff
@@ -965,6 +965,16 @@ CL_API_ENTRY cl_mem CL_API_CALL clCreateBufferIntelFPGA(cl_context context,
                                            errcode_ret);
 }
 
+// Create a buffer with property
```
pcolberg, line 968: Nitpick: how about the upstream description, `Create a buffer with additional properties`?

src/acl_mem.cpp
```diff
@@ -965,6 +965,16 @@ CL_API_ENTRY cl_mem CL_API_CALL clCreateBufferIntelFPGA(cl_context context,
...
+// It is equivalent to clCreateBufferWithPropertiesINTEL
```
pcolberg, line 969: Nitpick: This comment duplicates the content of the function body, so you could remove it if you like. It really does not matter nor apply here, but in general, not duplicating the more obvious facts in comments avoids comment and code drifting apart eventually. Omitting the obvious facts may also motivate to ponder a bit deeper about the value a comment could provide to the reader (which again does not apply here).

pcolberg (changes requested):
Thanks @sherry-yuan!

I don't see any reason not to merge this right away. @zibaiwan, what do you think?

src/acl_mem.cpp
sherry-yuan: Thanks I have changed the comment and removed the other unnecessary comment.

zibaiwan:
> Thanks @sherry-yuan!
> 
> I don't see any reason not to merge this right away. @zibaiwan, what do you think?

Thanks @sherry-yuan @pcolberg . I agree this PR can be merged right away.

zibaiwan (approved):

sherry-yuan:
Thank you all for reviewing @zibaiwan @pcolberg ! I agree too that there isn't any downside in merging this, as long we don't announce to the public we are at 3.0, it should be all good.

May I get someone to press the merge button for me? I do not have the permission to do so.

pcolberg (approved):

PR #57, opened 2022-01-11
Add OCL Core 3.0 clCreateImageWithProperties function

### LINKED ISSUE #90: Implement OpenCL 3.0 clCreateImageWithProperties
https://github.com/intel/fpga-runtime-for-opencl/blob/950f21dd079dfd55a473ba4122a4a9dca450e36f/include/CL/cl_icd.h#L154-L170

sherry-yuan:
As a part of the effort in upgrading opencl version

Currently the child function it calls receive no properties because as of now, user cannot create image with property.

Closes #90

### DIFF
--- a/src/acl_mem.cpp
+++ b/src/acl_mem.cpp
@@ -1778,6 +1778,16 @@ ACL_EXPORT CL_API_ENTRY cl_mem CL_API_CALL clCreateImage(
   return clCreateImageIntelFPGA(context, flags, image_format, image_desc,
                                 host_ptr, errcode_ret);
 }
+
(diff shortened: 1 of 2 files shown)


pcolberg:
Rebased onto main branch and updated unit test to use `assert(image)` instead of `CHECK(image)` since Klocwork does not recognize the latter as a null pointer check.

Closing for now since, as the [comment](https://github.com/intel/fpga-runtime-for-opencl/blob/6da6bcd86522ce3b392d859c41e35b29874feec4/src/acl_mem.cpp#L1786) explains, properties are not yet handled by `clCreateImageWithProperties()`, and we do not have demand for the feature.

PR #58, opened 2022-01-14
Remove `please` in messages and documentations

sherry-yuan:
Avoid using please in technical documentations according to https://www.sitepoint.com/making-demands-when-to-say-please-in-instructions-and-ctas/

Thanks to @dseynhae for pointing this out, and thanks @pcolberg for fixing other places!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ endif()
 
 # We exclusively support Ninja as the generator used to build the project.
 if(NOT CMAKE_GENERATOR STREQUAL "Ninja")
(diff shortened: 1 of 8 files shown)


pcolberg (approved):
Thanks @sherry-yuan, looks good. I think it is fine to change these messages since they are not part of the officially documented API, but we should keep in the back of mind that this might require adaptation of integration tests which explicitly test for specific error messages.

sherry-yuan:
> keep in the back of mind that this might require adaptation of integration tests which explicitly test for specific error messages.

I have notified the corresponding communication channel.

PR #60, opened 2022-01-17
Add documentation for running clang-format

sherry-yuan:
Document which clang version to run to avoid workflow failure, clang-format git integration as well installation instruction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,11 +126,37 @@ branch that is checked out in a separate worktree.
     git log --stat -p aaaaaa...bbbbbb
     ```
 
+## Polishing a change
+
+Prior to pushing your change, remember to run [clang-format] on the `.c`,
+`.cpp`, and `.h` files, since the [clang-format workflow] may fail otherwise.
+You can do so by running the following command.
+
+```
+./scripts/clang-format.sh --verbose
+```
+
+To match the exact formatting expected by the [clang-format version in the
+workflow], you will need to obtain [clang version 13.0.0] and add its `bin`
+directory to `PATH`. If you are using RHEL/CentOS or SLES/openSUSE, try
+installing one of the `clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-*.tar.xz`
+that is compatible with the libc and libstdc++ versions provided by your
+distribution.
+
+If you wish to format a committed or staged file, you can use clang-format git
+integration, [git-clang-format]. The main benefit of `git clang-format` is the
+ability to format a specific commit or only staged files.
+
 [Intel organization]: https://github.com/intel
 [Intel® FPGA Runtime for OpenCL™ Software Technology]: https://github.com/intel/fpga-runtime-for-opencl
+[clang version 13.0.0]: https://github.com/llvm/llvm-project/releases/tag/llvmorg-13.0.0
+[clang-format version in the workflow]: https://github.com/intel/fpga-runtime-for-opencl/blob/fc99b92704a466f7dc4d84bd45d465d64d03dbb0/container/ubuntu-20.04-clang/Dockerfile#L9
+[clang-format workflow]: https://github.com/intel/fpga-runtime-for-opencl/blob/fc99b92704a466f7dc4d84bd45d465d64d03dbb0/.github/workflows/clang-format.yml
+[clang-format]: https://clang.llvm.org/docs/ClangFormat.html
 [create a forked repository]: https://docs.github.com/en/get-started/quickstart/fork-a-repo#forking-a-repository
 [fork and pull model]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/about-collaborative-development-models#fork-and-pull-model
 [git worktree]: https://git-scm.com/docs/git-worktree
+[git-clang-format]: https://github.com/llvm/llvm-project/blob/9e634b35ff51d0eb2b38013111491e88bdbae388/clang/tools/clang-format/git-clang-format
 [pull requests]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests
 [remote repositories]: https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes
 [runtime repository]: https://github.com/intel/fpga-runtime-for-opencl
--- a/scripts/clang-format.sh
+++ b/scripts/clang-format.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+set -eu -o pipefail
+
+find . -regex '.*\.\(c\|cpp\|h\)' | sort | xargs clang-format -i "$@"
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -126,6 +126,19 @@ branch that is checked out in a separate worktree.
     git log --stat -p aaaaaa...bbbbbb
     ```
 
+## Polish a change
+
+Prior to pushing your change, remember to run clang-format on the `c/cpp` files, clang-format workflow may fail otherwise. You can do so by running the following command.
```
pcolberg, line 131: Try linking as much as possible; you never know whether the context might be valuable to a user. Here I would link `clang-format` to the clang-format website and `clang-format workflow` to our clang-format workflow file.

CONTRIBUTING.md
```diff
@@ -126,6 +126,19 @@ branch that is checked out in a separate worktree.
...
+
+`find . -regex '.*\.\(c\|cpp\|h\|hpp\)' | sort | xargs clang-format --verbose -i`.
```
pcolberg, line 133: This is the exact command from the workflow, but I am wondering if it makes sense to shorten it, e.g.,
```
find . -regex '.*\.\(c\|cpp\|h\)' -exec clang-format -i {} +
```
In the workflow the verbose listing of processed files does not disturb when not needed, but it does when manually used on the console. We don't actually have `hpp` files, maybe that should be removed from the workflow as well.

Nitpick: You can use a Markdown code block here:
~~~
```
code
```
~~~

CONTRIBUTING.md
```diff
@@ -126,6 +126,19 @@ branch that is checked out in a separate worktree.
...
+
+To do this, you will need clang version 13.0.0, which you can download from [llvm](https://github.com/llvm/llvm-project/releases/tag/llvmorg-13.0.0). Then add its bin directory to PATH.
+
+Note: If you are using Fedora, CentOS, or SUSE distribution, you can install `clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-xx.xx.tar.xz`.
+
+If you wish to format a committed or staged file. You can use clang-format git integration: [`git-clang-format`](https://github.com/llvm/llvm-project/blob/9e634b35ff51d0eb2b38013111491e88bdbae388/clang/tools/clang-format/git-clang-format).
```
pcolberg, line 139: Use a comma: `If you wish to format a committed or staged file, you can use`

pcolberg: What is the main benefit of using `git clang-format` over running `clang-format` manually?

pcolberg (changes requested):
Thanks @sherry-yuan, your documentation will make it much easier to pass the `clang-format` workflow on first try.

CONTRIBUTING.md
sherry-yuan: I gave it a try, and it seems like that will result in some version mismatch error.

CONTRIBUTING.md
sherry-yuan: The main benefit of `git clang-format` is the ability to format a specific commit without having to format, and fixup commits.
Or it can be used on only the staged files.

CONTRIBUTING.md
pcolberg: What error are you getting for the `-exec` variant?

In any case, how about adding a one-liner script for the longer clang-format script, e.g.,
```
find . -regex '.*\.\(c\|cpp\|h\)' | sort | xargs clang-format -i "$@"
```
which allows passing `--verbose` as an optional argument to list files if desired.

CONTRIBUTING.md
pcolberg: Thanks Sherry. In terms of practical workflows, is `git clang-format` always preferable over manually running `clang-format`? Is there a reason not to recommend `git clang-format`? (Sorry for asking these simple questions; I use the [vim-clang-format](https://github.com/rhysd/vim-clang-format) plugin for automatic formatting.)

CONTRIBUTING.md
sherry-yuan: It was an warning about version not found `clang-format: /lib64/libtinfo.so.5: no version information available (required by clang-format)` but overall the files are still getting formatted. I have changed the command to the one-liner you proposed.

CONTRIBUTING.md
sherry-yuan: I think it really depend on what situation the developer is in. I personally would like to solely rely on that single line command / or some sort of plugins of IDE. However if a developer accidentally committed a change without formating (maybe because forgot to formating) and wish to do it after in one line, then git-clang-format is the way to go. Further, if we need to import an upstream project in the future, and that project is not completely formated, then some sort of surgical formatting like git-clang-format will meet our need (to avoid noise in PR).

CONTRIBUTING.md
sherry-yuan: > how about adding a one-liner script for the longer clang-format script

Its a good idea. In this case, should we add for both linux and windows?

pcolberg: On Windows we require `cygwin` (which is used in some of the unit tests), so a bash script only should be fine.

sherry-yuan: Got it, I have added the bash script.

scripts/clang-format.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Install Intel FPGA SDK for OpenCL Pro Edition
+# https://fpgasoftware.intel.com/
+
+set -eu -o pipefail
+
+find . -regex '.*\.\(c\|cpp\|h\)' -exec clang-format -i "$@" {} +
```
pcolberg, line 10: Could you use the initial `| sort | xargs` variant and call this script in the clang-format workflow, so it gets tested?

Without `--verbose` the effect is the same, but with the optional `--verbose` the list of files is sorted.

scripts/clang-format.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Install Intel FPGA SDK for OpenCL Pro Edition
+# https://fpgasoftware.intel.com/
```
pcolberg, line 6: These two lines are not part of the license header used in the runtime.

scripts/clang-format.sh
sherry-yuan: I have changed it to the same command used in workflow, added it in workflow and added a test commit that test this script.

scripts/clang-format.sh
sherry-yuan: Thanks for catching that! now removed.

pcolberg (dismissed):

pcolberg:
Thanks @sherry-yuan. I cosmetically revised the Markdown to make it somewhat readable in raw format, but otherwise left everything as is. Good thinking on writing a temporary test of a failing `clang-format` run :slightly_smiling_face:

pcolberg:
Since I noticed you committed additions with message `fixup`: You can use `git commit --fixup=<commit>` to replicate the subject of an earlier commit, preceded by `fixup!`. When everything is ready for squashing, `git rebase -i --autosquash intel/main` automatically squashes these commits, discarding their messages.

pcolberg (approved):

sherry-yuan:
I see, thanks for the tip @pcolberg !

PR #61, opened 2022-01-18
Support early print buffer flushing in debug mode

zibaiwan:
Added a flow to flush printf() buffer before the Kernel ends. This will be helpful for debugging a hang Kernel. The Runtime periodically polls from the device and process the printf buffer from the last processed point.

If ACL_HAL_DEBUG = 1 and the Kernel is Hang, Kernel status (existing feature) and printf buffer will be flushed to stdout every 10 seconds (approximately). The thread will run if there are spare cycles.

If ACL_HAL_DEBUG = 3, Kernel status and printf buffer will flushed every 10 seconds, even the Kernel isn't hung.

The behavior of printf() is not altered in non-debug mode.

## Sample output

```
  Kernel  0 Status: 0x00048000 running:: Accelerator 0 printf buffer size is 160.
:: Calling acl_process_printf_buffer_fn with activation_id=1 and printf_size=160.


Previously processed buffer size is 0 
Another test string 14
This is a test
Another test string 11
Another test string 12
Another test string 13
  Kernel  0 Status: 0x00048000 running:: Accelerator 0 printf buffer size is 160.
:: Calling acl_process_printf_buffer_fn with activation_id=1 and printf_size=160.


Previously processed buffer size is 160 
All Printf() buffer has already been dumped
```

## ToDo
- [x] Fix Klocwork if applicable
- [x] Run Clang format
- [x] manual testing
- [x] Clear commit messages and Rebase
- [x] Run existing regtest to ensure existing feature still passes
- [ ] Create a new regtest to test this specific feature

### DIFF
--- a/include/acl_types.h
+++ b/include/acl_types.h
@@ -922,6 +922,10 @@ typedef struct _cl_kernel {
   // Eventually this should be an array of ACLDeviceBinaries similar
   // to how cl_program contains an array of dev_prog.
   const acl_device_binary_t *dev_bin;
+
+  // In ACL_HAL_DEBUG mode, printf buffer could be dumped before Kernel ends
+  // Therefore, we need to keep track of how much data has been processed.
+  size_t processed_printf_buffer_size;
 } _cl_kernel;
 
 ACL_DECLARE_CL_OBJECT_ALLOC_FUNCTIONS(cl_kernel);
@@ -1346,6 +1350,10 @@ typedef struct {
   // it sees the CL_COMPLETE status for the first time, but it won't
   // change after that.
   cl_uint num_printf_bytes_pending;
+
+  // Indicate whether this operation is dumping printf buffer before the Kernel
+  // for debug purpose
+  int debug_dump_printf = 0;
 } acl_device_op_info_t;
 
 // An operation to be performed on a device.
--- a/src/acl_printf.cpp
+++ b/src/acl_printf.cpp
@@ -38,8 +38,8 @@
 #pragma GCC visibility push(protected)
 #endif
 
-static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
-                                 unsigned size);
+static size_t l_dump_printf_buffer(cl_event event, cl_kernel kernel,
+                                   unsigned size);
 static void decode_string(std::string &print_data);
 
 // stores to global memory are bound by smallest global memory bandwidth we
@@ -719,12 +719,13 @@ static std::string::const_iterator get_data_elem_at_offset(
   return end_of_string;
 }
 
-static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
-                                 unsigned size) {
+static size_t l_dump_printf_buffer(cl_event event, cl_kernel kernel,
+                                   unsigned size) {
   unsigned global_offset;        // the location in the printf buffer
   unsigned single_printf_offset; // the offset of a single printf
   void (*hal_dma_fn)(cl_event, const void *, void *, size_t) = 0;
   int src_on_host = 1;
+  size_t dumped_buffer_size = 0;
 #ifdef _WIN32
   __declspec(align(64)) char
       buffer[ACL_PRINTF_BUFFER_TOTAL_SIZE]; // Aligned to 64, for dma transfers
@@ -743,11 +744,11 @@ static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
   if (!verify_types()) {
     printf("Host data types are incompatible with ACL compiler, ignoring "
            "printfs...\n");
-    return;
+    return dumped_buffer_size;
   }
 
   if (printf_infos.empty())
-    return;
+    return dumped_buffer_size;
 
   // Memory is on the device if all of these are true:
   //   The memory is not SVM or the device does not support SVM.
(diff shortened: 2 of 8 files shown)


src/acl_printf.cpp
```diff
@@ -769,8 +770,24 @@ static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
     hal_dma_fn = acl_get_hal()->copy_globalmem_to_hostmem;
   }
 
-  hal_dma_fn(NULL, kernel->printf_device_buffer->block_allocation->range.begin,
-             buffer, size);
+  // It needs the context from ACL_HAL_DEBUG instead of ACL_DEBUG
+  if (acl_get_hal()->get_debug_verbosity &&
+      acl_get_hal()->get_debug_verbosity() > 0) {
+    printf("Previously processed buffer size is %d \n", kernel->processed_printf_buffer_size);
```
pcolberg, line 776: For `size_t` the printf modifier is `%zu`.

include/acl_types.h
```diff
@@ -1346,6 +1350,10 @@ typedef struct {
   // it sees the CL_COMPLETE status for the first time, but it won't
   // change after that.
   cl_uint num_printf_bytes_pending;
+
+  // Indicate whether this operation is dumping printf buffer before the Kernel
+  // for debug purpose
+  int debug_dump_printf = 0;
```
pcolberg, line 1356: Would anything speak against making this a `bool`, since it only has two states?

src/acl_printf.cpp
```diff
@@ -769,8 +770,28 @@ static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
...
+           kernel->processed_printf_buffer_size);
+  }
+
+  // Check if we have already processed all the printf buffer
+  if (size > (unsigned int)kernel->processed_printf_buffer_size) {
+    void *unprocessed_begin = (void *)((char *)kernel->printf_device_buffer
+                                           ->block_allocation->range.begin +
+                                       kernel->processed_printf_buffer_size);
+    hal_dma_fn(NULL, unprocessed_begin, buffer,
+               size - kernel->processed_printf_buffer_size);
```
pcolberg, line 786: How about moving the assignment to `dumped_buffer_size` before this call and using it here?

pcolberg: `assert(size >= kernel->processed_printf_buffer_size)` might be useful to catch a possible underflow.

src/acl_printf.cpp
```diff
@@ -960,13 +981,14 @@ static void l_dump_printf_buffer(cl_event event, cl_kernel kernel,
...
 #endif
+  return dumped_buffer_size;
 }
 
 //
 // Schedule enqueue read buffer to read printf buffer
 // The activation ID is the device op ID.
 void acl_schedule_printf_buffer_pickup(int activation_id, int size,
-                                       int stalled) {
+                                       int debug_dump_printf) {
```
pcolberg, line 991: Does this HAL API also need to be updated?
https://github.com/intel/fpga-runtime-for-opencl/blob/67f7f880f7af5af0c1797c4dd2a70b4e27ba85ca/include/acl_hal.h#L65-L66

include/acl_types.h
pcolberg: Never mind, the type `int` is aligned with the HAL API, correct?
https://github.com/intel/fpga-runtime-for-opencl/blob/67f7f880f7af5af0c1797c4dd2a70b4e27ba85ca/include/acl_hal.h#L65-L66

src/acl_printf.cpp
```diff
@@ -980,17 +1002,21 @@ void acl_schedule_printf_buffer_pickup(int activation_id, int size,
...
     // This address is stable, given a fixed activation_id.
     // So we don't run into race conditions.
     acl_device_op_t *op = doq->op + activation_id;
-
-    stalled = stalled; // this argument is no longer used!
-
     op->info.num_printf_bytes_pending = (cl_uint)size;
+
+    // Propagate the operation info
+    if (debug_dump_printf == 1) {
```
pcolberg, line 1015: What happens if `debug_dump_printf` is any other non-zero value, say 2?

src/acl_printf.cpp
```diff
@@ -980,17 +1002,21 @@ void acl_schedule_printf_buffer_pickup(int activation_id, int size,
...
+      op->info.debug_dump_printf = 1;
+    } else {
+      op->info.debug_dump_printf = 0;
+    }
```
pcolberg, line 1019: You could shorten to
```
op->info.debug_dump_printf = debug_dump_printf ? 1 : 0;
```
i.e., zero input value zero yields 0 and non-zero input value is normalized to 1.

src/acl_printf.cpp
```diff
@@ -980,17 +1002,21 @@ void acl_schedule_printf_buffer_pickup(int activation_id, int size,
...
   fflush(stdout);
 #endif
   if (activation_id >= 0 && activation_id < doq->max_ops) {
     // This address is stable, given a fixed activation_id.
     // So we don't run into race conditions.
     acl_device_op_t *op = doq->op + activation_id;
-
-    stalled = stalled; // this argument is no longer used!
-
     op->info.num_printf_bytes_pending = (cl_uint)size;
```
pcolberg, line 1012: Why is `num_printf_bytes_pending` not `size_t`?

pcolberg (changes requested):
Thanks @zibaiwan! Could you include a commit message that summarizes the change and also motivates why it was made?

Have you taken a look at the unit tests to see if anything could be inserted to test this new functionality, as far as is feasible in a unit test? For instance, checking that the new state variables added here correspond to the expected state.

src/acl_printf.cpp
pcolberg: Please never mind, this is not part of the change.

src/acl_printf.cpp
zibaiwan: Thanks Peter. Your suggestion looks cleaner and handled any other non-zero value.

zibaiwan:
@pcolberg , thank you very much for your review. I have updated the PR based on your suggestions.

src/acl_printf.cpp
```diff
@@ -1009,7 +1031,15 @@ void acl_process_printf_buffer(void *user_data, acl_device_op_t *op) {
 
     // Grab the printf data and emit it.
     cl_uint num_bytes = op->info.num_printf_bytes_pending;
-    l_dump_printf_buffer(event, kernel, num_bytes);
+    size_t dumped_buffer_size = l_dump_printf_buffer(event, kernel, num_bytes);
+
+    if (op->info.debug_dump_printf == 1) {
```
pcolberg, line 1036: Since `int debug_dump_printf` has boolean semantics:
```c++
     if (op->info.debug_dump_printf) {
```

src/acl_printf.cpp
```diff
@@ -1021,8 +1051,13 @@ void acl_process_printf_buffer(void *user_data, acl_device_op_t *op) {
     acl_memory_barrier();
 
     // Allow the kernel to continue running.
-    acl_get_hal()->unstall_kernel(
-        event->cmd.info.ndrange_kernel.device->def.physical_device_id, op->id);
+    // We don't need to unstall the kernel during the early flushing during
+    // debug.
+    if (op->info.debug_dump_printf == 0) {
```
pcolberg, line 1056:
```c++
    if (!op->info.debug_dump_printf) {
```

pcolberg (changes requested):
Thanks @zibaiwan for the additional tests, much appreciated :slightly_smiling_face:

sherry-yuan:
A small question about a (maybe) corner case: What happens if the kernel stalls before the next 10 sec printf buffer dumps?

zibaiwan:
> A small question about a (maybe) corner case: What happens if the kernel stalls before the next 10 sec printf buffer dumps?

Good question. If a kernel is stalled because the printf buffer is full (runtime doesn't seem to stall for other reason, even it does, it wouldn't matter here), it will print the rest of the printf buffer which hasn't been dumped by the debug dumps.

src/acl_printf.cpp
zibaiwan: Thanks for the suggestion!

pcolberg (approved):

PR #62, opened 2022-01-19
Sync Khronos OpenCL-Header cl_intel_mem_alloc_buffer_location extension

sherry-yuan:
--------------------------------------------------------------------------------

Done to align with buffer location opencl extension in https://github.com/KhronosGroup/OpenCL-Headers/blob/80c10b1f65b932894b830da7cd37bc56c541bae4/CL/cl_ext.h#L2350-L2361

This extension allows user to specify which device global memory the memory should be allocated to.


pcolberg:
Great upstream work, Sherry :slightly_smiling_face: 

Instead of copying only the new `cl_intel_mem_alloc_buffer_location` section, would you like to update the OpenCL headers to the latest version containing your change? We want to ensure that these are an exact copy of upstream at any time.

sherry-yuan:
> would you like to update the OpenCL headers to the latest version containing your change?

Sure! I will sync everything.

sherry-yuan:
The sync is complete.

A caveat: one of the files we currently have does not exist in khronos registry. 
for example: https://github.com/intel/fpga-runtime-for-opencl/blob/main/include/CL/cl_ext_intelfpga.h

I feel like we might not want to push these to khronos, but at the same time we do not want to have our own enum that potentially will conflict with future khronos headers. Things in this file need to be updated periodically.

pcolberg:
Thanks Sherry. Could you set the modes of the files back to non-executable (`chmod -x *`)?

include/CL/cl_version.h
```diff
@@ -19,7 +19,8 @@
 
 /* Detect which version to target */
 #if !defined(CL_TARGET_OPENCL_VERSION)
-#define CL_TARGET_OPENCL_VERSION 120
+#pragma message("cl_version.h: CL_TARGET_OPENCL_VERSION is not defined. Defaulting to 300 (OpenCL 3.0)")
+#define CL_TARGET_OPENCL_VERSION 300
```
pcolberg, line 23: Could you briefly note this particular change in the commit message, and why it won't break anything?

sherry-yuan: Added in commit: Upgrade default version to opencl core 3.0. This will not break existing functio
nality because we are currently already building with 3.0 (see cMakeList). All f
uture version of opencl core is backward compatible.

pcolberg:
Thanks @sherry-yuan! I pushed a minor revision that reverts the executable bit on `include/CL/.clang-format` and mentions the source commit of the files, https://github.com/KhronosGroup/OpenCL-Headers/commit/80c10b1f65b932894b830da7cd37bc56c541bae4.

pcolberg (approved):

Issue #38, opened 2021-12-06
Add additional kernel status update case


sherry-yuan:
A follow-up to #37

Kernel interrupt are sent when kernel changed status from init to running, however this case was not checked in the signal handler
https://github.com/intel/fpga-runtime-for-opencl/blob/df4be1924268cdb7841da2a6b0618b7bb8a47628/src/acl_kernel_if.cpp#L1257

And previously it was assumed that the finish bit must be set after checking "all other possible cases" (i.e buffer full, kernel finish, kernel stall, profiler interrupt). This assumption was enforced by an assert statement which was removed in #37

The full solution is to add additional checking on status register's values so that we don't reach the final "kernel finish" case checking without checking for kernel running status.

sherry-yuan:
Confirmed the assert should be removed, there are two flow that will call on acl_kernel_if_update_status: polling and interrupt flow . For polling, there is no guarantee that >0kernel will finish when runtime polls for status.

sherry-yuan:
Realized anything that is not done, stall or profiler update will not reach assert.
https://github.com/intel/fpga-runtime-for-opencl/blob/df4be1924268cdb7841da2a6b0618b7bb8a47628/src/acl_kernel_if.cpp#L1335-L1338

However there are still cases where some kernel in queue is done but finish count is 0, log below;
```
:: Accelerator 11 reporting status 2.
:: Accelerator 11 is done.
HAL Reading from Kernel: 4 bytes 1628 -> 91c1bff2d8
108360416261 acl_kernel_if_write_32b 00000020 0
108360416264 acl_kernel_if_read_32b 00001628 0
HAL Reading from Kernel: 4 bytes 20 -> 91c14f9fd0
:: Accelerator 11 has 0 finishes.
```
Normally it should look like (always like this with no other writes in between):
```
:: Accelerator 11 reporting status 2.
:: Accelerator 11 is done.
HAL Reading from Kernel: 4 bytes 1628 -> 91c1bff2d8
108360062459 acl_kernel_if_read_32b 00001628 1
:: Accelerator 11 has 1 finishes.
```

This suggests writing 0 at CSR address 00000020 has overwritten values on 00001628.

The writing 0 at CSR address 00000020 only happens closely after launching a new kernel.  

x20 address corresponds to OFFSET_KERNEL_CRA_SEGMENT
[...]

PR #63, opened 2022-01-19
Add Thread Sanitizer CI/CD

sherry-yuan:
ThreadSanitizer is a tool that detects data races. It consists of a compiler instrumentation module and a run-time library.

This help us to identify race conditions within the program. It will have greater coverage if we can mock MMD.

Addresses #38

### DIFF
--- a/.github/workflows/tsan.yml
+++ b/.github/workflows/tsan.yml
@@ -1,0 +1,99 @@
+# Copyright (C) 2021 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+
+name: thread sanitizer
+
+# https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/
+permissions:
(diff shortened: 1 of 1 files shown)


sherry-yuan:
Apparently tsan doesn't work on internal workstations because both the OS and gcc tsan is too old. 
> So the immediate cause of this problem is that the lsan memory allocator in gcc 8 uses a fixed address of 0x600000000000ULL for starting its memory maps.  This is beyond the end of virtual memory for some older kernels (which end at 0x400000000000ULL) but within range for newer ones.

Have to test the flow on using provided images.

sherry-yuan:
Currently this depend on https://github.com/intel/fpga-runtime-for-opencl/issues/71 being fixed for third party libraries, I currently need to focus on a release critical feature before I could get to fixing third party libraries.

pcolberg:
Thanks @sherry-yuan! Small clarification, `acl_threadsupport` is a first-party library.

pcolberg (approved):

sherry-yuan:
Thanks @pcolberg :)

PR #65, opened 2022-01-24
Device global implementation

sherry-yuan:
Currently the device global pointer is being mocked by passing an actual device allocation into the function. However this should be removed after the pointer is available through autodiscovery string.


src/acl_mem.cpp
```diff
@@ -405,6 +405,88 @@ int acl_bind_buffer_to_device(cl_device_id device, cl_mem mem) {
...
+    const void* dev_global_ptr) {
+  cl_int status;
+  cl_kernel kernel = clCreateKernelIntelFPGA(program, name, &status);
+  if (status != CL_SUCCESS) {
+    return status;
+  }
+  // TODO: get dev_global_ptr from autodiscovery instead later
+  // Assuming below returns same thing as `clDeviceMemAllocINTEL`, where exactly is dev global ptr being read from?
+  // What format is the read dev global, is it a pointer just like the return value of `clDeviceMemAllocINTEL` or is it something else?
+  // void* dev_global_ptr = kernel->dev_bin->get_devdef().autodiscovery_def.?
```
sherry-yuan, line 452: What exact format is being read from autodiscovery? Is it exactly the same as the return of `clDeviceMemAllocINTEL`?

src/acl_mem.cpp
```diff
@@ -405,6 +405,88 @@ int acl_bind_buffer_to_device(cl_device_id device, cl_mem mem) {
...
+
+  // do we support ptr being a buffer instead of usm pointer? it seems to be the case on spec
```
sherry-yuan, line 454: Do we support `ptr` being buffer also? Or is it just USM?

src/acl_mem.cpp
```diff
@@ -405,6 +405,88 @@ int acl_bind_buffer_to_device(cl_device_id device, cl_mem mem) {
...
+  status = clSetKernelArgMemPointerINTEL(kernel, 0, ptr);
+  if (status != CL_SUCCESS) {
+    return status;
+  }
+  // should kernel header contain offset or not?
+  // offset is always byte offset?
```
sherry-yuan, line 460: Should kernel take the offset information, or should the passed in dev pointer be offset already? If so, is offset always byte offset (I recall it is, but just need to confirm)?

src/acl_mem.cpp
```diff
@@ -405,6 +405,88 @@ int acl_bind_buffer_to_device(cl_device_id device, cl_mem mem) {
...
+  // When passing dev global pointer directly, make sure REMOVE_VALID_CHECKS is defined.
+  // Otherwise this ptr is not existing in context -> cause checks to fail
+  // Given kernel arg must be a deivce usm pointer: When ptr is a host/shared usm pointer, this function is expected to copy it to device first?
```
sherry-yuan, line 463: In the planned kernel header, it mentioned the arg is device pointer. In this case, I assume all passed in pointer need to be transformed to device first. i.e if `ptr` is a host/shared usm pointer, this function should create a device usm allocation and copy data from host to device?

test/acl_mem_test.cpp
```diff
@@ -218,6 +218,318 @@ MT_TEST_GROUP(acl_mem) {
...
+      // status = clEnqueueWriteBuffer(m_cq, mem1, CL_FALSE, 0, strsize, str, 0, 0,
+      //                               &write_event);
+      // CHECK_EQUAL(CL_SUCCESS, status);
+      // status = clEnqueueReadGlobalVariableINTEL(m_cq, m_program, "kernel15_dev_global", CL_FALSE, strsize, 0, (void*)(&mem2), 1, &write_event, &copy_event, &mem1);
+      
+      // usm version
+      status = clEnqueueReadGlobalVariableINTEL(m_cq, m_program, "kernel15_dev_global", CL_FALSE, strsize, 0, src_ptr, 0, NULL, &copy_event, dev_global_ptr);
+      CHECK_EQUAL(CL_SUCCESS, status);
+      status = clEnqueueReadBuffer(m_cq, mem2, CL_FALSE, 0, strsize, resultbuf, 1,
+                                    &copy_event, &read_event);
```
sherry-yuan, line 493: Is this the expected way of using it?

artemrad:
What is the purpose of the number field in each at each index of the autodiscovery string? 

I would also think about how to add fields containing information about properties associated with each device_global.

sherry-yuan:
Thanks @artemrad for the comment!

> What is the purpose of the number field in each at each index of the autodiscovery string?

num_dev_global [(**num_dev_global_field_1**, dev_global_name_1, dev_global_address_1, dev_global_size_1), (**num_dev_global_field_2**, dev_global_name_2, dev_global_address_2, dev_global_size_2), (...), ...]

By number field you mean the bolded part of above?

The `num_dev_global_field` keep track of the number of properties associated with each device global. We need this number because the autodiscovery parser relies counting number of remaining item to read per device global. Further, this setup ensure forward compatibility. i.e if we decided to add additional properties in the future for some specific device global, we can easily do so by increasing the `num_dev_global_field` (which tells the parser there are one more field to read for that specific device global).

> I would also think about how to add fields containing information about properties associated with each device_global.

I think the above mechanism makes sure we can add additional properties for each device global while maintaining forward compatibility? Currently the attributes include `dev_global_name`, `dev_global_address`, `dev_global_size` (there can always be more, let me know if I am missing something). Correct me if I understand `attribute` wrong.

sherry-yuan:
Update milestone to 2022.4 as that is the current target.

src/acl_mem.cpp
```diff
@@ -405,6 +406,339 @@ int acl_bind_buffer_to_device(cl_device_id device, cl_mem mem) {
...
+    cl_event *event) {
+  cl_int status;
+
+  // Get copy kernel from the program
+  cl_kernel kernel =
+      clCreateKernelIntelFPGA(program, "copy_device_global", &status);
+  if (status != CL_SUCCESS) {
+    return status;
+  }
+
```
pcolberg, line 448: You can insert an `assert(kernel)` after checking the status to appease Klocwork.

test/acl_usm_test.cpp
```diff
@@ -1206,4 +1244,74 @@ MT_TEST(acl_usm, memfill_usm) {
...
+
+  // Prepare host memory
+  syncThreads();
+  // Host pointer example
+  void *src_ptr = malloc(strsize);
+  CHECK(src_ptr != NULL);
+  // Host USM example
+  // void* src_ptr = clHostMemAllocINTEL(m_context, NULL, strsize, 1, &status);
+  // CHECK_EQUAL(CL_SUCCESS, status);
+  // CHECK(src_ptr != NULL);
```
pcolberg, line 1265: Commented code should be removed; unless you are intend to enable this before merge?

sherry-yuan:
Autodiscovery change is separated out into a separate PR (because the full implementation depend on updating upstream khronos header, which depend on the full implementation to be ready in sycl runtime). Autodiscovery change could be merged first

Autodiscovery change PR: https://github.com/intel/fpga-runtime-for-opencl/pull/100

PR #68, opened 2022-01-28
Install OpenCL header CL/cl_layer.h

pcolberg:
This amends https://github.com/intel/fpga-runtime-for-opencl/pull/62

OpenCL layers are an experimental feature which we don't use, but there
is no downside to installing the header for the sake of completeness.

https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/cd7d07cfa667d8d959b4272be45cf217a65c2948/README.md#about-layers


zibaiwan (approved):
Looks good! Thanks Peter!

PR #69, opened 2022-02-02
Load libz.so.1

alisanikiforova:
Many OSs have only libz.so.1 but no libz.so, so when loading zlib, want to check both and load libz.so.1 first, and then double check libz.so if that fails.

### DIFF
--- a/lib/pkg_editor/src/zlib.c
+++ b/lib/pkg_editor/src/zlib.c
@@ -86,6 +86,10 @@ static void *zlib_load_library() {
   void *const library = (void *)LoadLibraryA(WINDOWS_ZLIB_PATH);
   return library;
 #else
+  void *const dlopen_handle = dlopen("libz.so.1", RTLD_NOW);
+  if (dlopen_handle) {
+    return dlopen_handle;
+  }
   return dlopen("libz.so", RTLD_NOW);
 #endif
 }


zibaiwan (changes requested):
@alisanikiforova, thanks for fixing this. Can you please squash those two commits into one commit and make the commit message structured like Title + Content? Thanks! Let me know if you need any help with squashing the commits. I usually use `git reset --soft HEAD~2 &&
git commit` and you probably need to push with force. You can also use 'git rebase -i'.

lib/pkg_editor/src/zlib.c
```diff
@@ -86,7 +86,12 @@ static void *zlib_load_library() {
   void *const library = (void *)LoadLibraryA(WINDOWS_ZLIB_PATH);
   return library;
 #else
-  return dlopen("libz.so", RTLD_NOW);
+  void *dlopen_handle = (void *)dlopen("libz.so.1", RTLD_NOW);
```
pcolberg, line 89: [`dlopen(3)`](https://man7.org/linux/man-pages/man3/dlopen.3.html) returns a `void *` so you can omit the cast. The underlying idea is to only use casts when needed, and scrutinize the few casts that are indeed needed, especially C-style casts since these easily lead to undefined behaviour.

zibaiwan (approved):

lib/pkg_editor/src/zlib.c
```diff
@@ -86,7 +86,12 @@ static void *zlib_load_library() {
...
+  if (dlopen_handle == NULL) {
+    return dlopen("libz.so", RTLD_NOW);
+  } else {
+    return dlopen_handle;
+  }
```
pcolberg, line 94: How about inverting the condition to make it slightly easier to follow the control flow?
```c++
void *const dlopen_handle = dlopen(...);
if (dlopen_handle) {
  return dlopen_handle;
}
return dlopen(...);
```
The early-return pattern is especially useful for more than two cases to avoid deep nesting of conditionals.

pcolberg: (Applying our recent C++ training :slightly_smiling_face:, in C++17 the variable can be scoped to the conditional
```c++
if (void *const dlopen_handle = dlopen(...); dlopen_handle) {
  return dlopen_handle;
}
return dlopen(...);
```
but the runtime is currently built with C++11 where that construct is not available.)

pcolberg (changes requested):
Thanks @alisanikiforova! Just a few minor cosmetic suggestions:

pcolberg (approved):

zibaiwan (approved):

PR #70, opened 2022-02-03
Document git identity setup

pcolberg:
This is well known but it does not hurt to briefly mention it in any case.


zibaiwan (approved):
Thanks Peter! This is worth noting.

PR #72, opened 2022-02-04
Migrate from CentOS to Rocky Linux

pcolberg:
CentOS 8 was discontinued and replaced by a rolling, continuously
updated distribution, CentOS stream that is always ahead of RHEL 8.

The Intel oneAPI Base Toolkit supports Rocky Linux 8 instead.

https://blog.centos.org/2020/12/future-is-centos-stream/
https://www.intel.com/content/www/us/en/developer/articles/system-requirements/intel-oneapi-base-toolkit-system-requirements.html

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       matrix:
         container:
-          - centos-8-dev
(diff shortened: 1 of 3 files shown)


zibaiwan (approved):
Thanks @pcolberg !

PR #73, opened 2022-02-04
Build and test on Rocky Linux

pcolberg:
This depends on https://github.com/intel/fpga-runtime-for-opencl/pull/72


aditikum (approved):
Thanks @pcolberg! LGTM

PR #75, opened 2022-02-07
Upgrade to c++17 standard

sherry-yuan:
c++17 supports many more features that allows greater flexibilities, and potential performance optimization.

https://en.cppreference.com/w/cpp/17

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 3.10)
 project(fpga-runtime-for-opencl C CXX)
 
 # Set default C++ standard property of targets.
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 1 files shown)


pcolberg (changes requested):
Thanks @sherry-yuan!

Could you extend the commit message with a quick summary of compiler support? You can use the C++17 core and library feature tables in https://en.cppreference.com/w/cpp/compiler_support to figure out the minimum required compiler versions for GCC and MSVC. Ubuntu 18.04 has GCC 7.5 and for Windows we support MSVC 2017.

sherry-yuan:
> Could you extend the commit message with a quick summary of compiler support?

Thanks for the suggestion! I checked the compiler versions and we satisfied all the requirements (the minimum required to access all feature is 19.14, and we are at 19.16, phew, that was close!)

I put how to check MSVC compiler version in commit number too.

pcolberg (approved):
Perfect, thanks @sherry-yuan!

Issue #71, opened 2022-02-04
TSAN Detected Data Race: Fix or Note as False Positive


sherry-yuan:
The following are list of [tsan detected data races](https://github.com/intel/fpga-runtime-for-opencl/pull/63)

The following happened in threadsupport package (which we should note as false positive, given we don't want to change these third party libraries.

```
WARNING: ThreadSanitizer: data race (pid=714)
  Write of size 4 at 0x559b36f693d4 by thread T4 (mutexes: write M53):
    #0 test_thread(void*) ../lib/acl_threadsupport/test/acl_threadsupport_test.cpp:105 (acl_threadsupport_test+0xa490)

  Previous read of size 4 at 0x559b36f693d4 by main thread:
    #0 threadsupport_threads_Test::testBody() ../lib/acl_threadsupport/test/acl_threadsupport_test.cpp:128 (acl_threadsupport_test+0xaac8)
    #1 Utest::executePlatformSpecificTestBody() ../lib/CppUTest/src/UtestPlatformGcc.cpp:82 (acl_threadsupport_test+0x1d278)
    #2 Utest::runInThread() ../lib/CppUTest/src/Utest.cpp:99 (acl_threadsupport_test+0x1aa94)
    #3 threadsupport_threads_Test::runInThreadOnCopy() <null> (acl_threadsupport_test+0x108ef)
    #4 Utest::executePlatformSpecificRunInThreads() ../lib/CppUTest/src/UtestPlatformGcc.cpp:175 (acl_threadsupport_test+0x1d7aa)
    #5 Utest::run(TestResult&) ../lib/CppUTest/src/Utest.cpp:83 (acl_threadsupport_test+0x1a9e7)
    #6 Utest::runOneTest(TestPlugin*, TestResult&) ../lib/CppUTest/src/Utest.cpp:69 (acl_threadsupport_test+0x1a8f6)
    #7 Utest::executePlatformSpecificRunOneTest(TestPlugin*, TestResult&) ../lib/CppUTest/src/UtestPlatformGcc.cpp:100 (acl_threadsupport_test+0x1d4ba)
    #8 Utest::runOneTestWithPlugins(TestPlugin*, TestResult&) ../lib/CppUTest/src/Utest.cpp:63 (acl_threadsupport_test+0x1a873)
    #9 TestRegistry::runAllTests(TestResult&) ../lib/CppUTest/src/TestRegistry.cpp:72 (acl_threadsupport_test+0x19c4d)
[...]

PR #76, opened 2022-02-08
Add thread sanitizer option to cmake

sherry-yuan:
----------------------------------------------

Thread sanitizer (TSAN) is a tool that detect data races, deadlocks  between threads. (it also sometimes report heap-use-after-free in same thread check #71 for issues reported in current repo)

For more information, checkout https://clang.llvm.org/docs/ThreadSanitizer.html

Addresses #38

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,7 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
 endif()
 
 # https://clang.llvm.org/docs/AddressSanitizer.html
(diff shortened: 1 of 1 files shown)


pcolberg:
Thanks @sherry-yuan! I revised the CMake code slightly to align with the existing `ACL_WITH_ASAN` option. We don't currently test with Clang, but it is not too much effort to keep this option open for the future. The `C` and `CXX` compilers may theoretically be from a different vendor, so for correctness the compiler ID should be checked separately for each language.

pcolberg (approved):

sherry-yuan:
Thanks @pcolberg, will do for future compiler checks!

PR #84, opened 2022-02-18
acl_hash: fix signedness conversion warnings for internal tail buffer

pcolberg:
The signedness of `char` is platform-dependent, e.g., signed on the
platforms supported by the runtime, which causes the following integer
conversion warnings. `char` is appropriate for a buffer of characters,
whereas `unsigned char` is appropriate for a buffer of arbitrary bytes.

The signedness change only affects the internal implementation, i.e. the
internal tail buffer and the internal functions using that buffer. The
unit tests provide sufficient coverage to ensure correctness.

```
[2/14] Building C object lib/acl_hash/CMakeFiles/acl_hash.dir/src/acl_hash.c.o
../lib/acl_hash/src/acl_hash.c: In function ‘l_close’:
../lib/acl_hash/src/acl_hash.c:189:14: warning: conversion to ‘char’ alters ‘int’ constant value [-Wconversion]
   extra[0] = 0x80; // append '1' bit. Assume big endian even within byte.
              ^~~~
../lib/acl_hash/src/acl_hash.c:199:20: warning: conversion to ‘char’ from ‘uint64_t {aka long unsigned int}’ may alter its value [-Wconversion]
     extra[7 - i] = user_bits & 0xff;
                    ^~~~~~~~~
```

Prerequisite of https://github.com/intel/fpga-runtime-for-opencl/issues/85

### DIFF
--- a/lib/acl_hash/include/acl_hash/acl_hash.h
+++ b/lib/acl_hash/include/acl_hash/acl_hash.h
@@ -55,7 +55,7 @@ typedef struct {
   uint32_t h0, h1, h2, h3, h4;
   // The (len % 64) most recent bytes to be added.
   // tail[0] is the earliest data byte, and so on.
-  char tail[64];
+  unsigned char tail[64];
 } acl_hash_sha1_context_t;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


zibaiwan (approved):
Thanks Peter for fixing this warning!

PR #88, opened 2022-03-02
Check if device support heterogeneous memory

sherry-yuan:
Currently sycl runtime is trying to distinguish whether buffer location is supported on the target device using lower level runtime getDeviceInfo call to determine it. However current runtime get device info does not yet support accepting the buffer location property yet, this change addes it.

The referenced sycl runtime PR: https://github.com/intel/llvm/pull/5604

This change is only needed if sycl runtime cannot come up with an workaround.

### DIFF
--- a/test/acl_device_test.cpp
+++ b/test/acl_device_test.cpp
@@ -324,10 +324,7 @@ MT_TEST(DeviceInfo, basic) {
     CHECK_EQUAL(CL_SUCCESS, clGetDeviceInfo(device, queries[i], str_size,
                                             &str[0], &size_ret));
 
-    if (queries[i] != CL_DEVICE_EXTENSIONS) {
-      // Non-empty result for most supported queries.
-      CHECK(size_ret > 0);
-    }
+    CHECK(size_ret > 0);
 
     switch (queries[i]) {
     case CL_DEVICE_VENDOR:
--- a/test/acl_device_test.cpp
+++ b/test/acl_device_test.cpp
@@ -14,6 +14,7 @@
 
 #include <acl.h>
 #include <acl_context.h>
+#include <acl_platform.h>
 #include <acl_types.h>
 #include <acl_util.h>
 
@@ -325,6 +326,10 @@ MT_TEST(DeviceInfo, basic) {
                                             &str[0], &size_ret));
 
     CHECK(size_ret > 0);
+
+    if (queries[i] == CL_DEVICE_EXTENSIONS) {
+      CHECK(strcmp(str, acl_platform_extensions()) == 0);
+    }
 
     switch (queries[i]) {
     case CL_DEVICE_VENDOR:
--- a/include/acl_platform.h
+++ b/include/acl_platform.h
@@ -21,6 +21,7 @@ extern int platform_owner_tid;
 void acl_init_platform(void);
 void acl_finalize_init_platform(unsigned int num_devices,
                                 const cl_device_id *devices);
+const char *acl_platform_extensions(void);
 
 #if defined(__cplusplus)
 } /* extern "C" */
--- a/src/acl_platform.cpp
+++ b/src/acl_platform.cpp
@@ -257,6 +257,34 @@ int acl_platform_is_valid(cl_platform_id platform) {
   return platform == &acl_platform;
 }
 
+const char *acl_platform_extensions() {
+  return "cl_khr_byte_addressable_store" // yes, we have correct access to
+                                         // individual bytes!
+         " cles_khr_int64" // yes, we support 64-bit ints in the embedded
+                           // profile
+
+         " cl_khr_icd"
+#if ACL_SUPPORT_IMAGES == 1
+         " cl_khr_3d_image_writes"
+#endif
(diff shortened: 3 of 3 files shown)


sherry-yuan:
It seems like there is a workaround in sycl runtime: https://github.com/intel/llvm/pull/5604#discussion_r818011979

test/acl_device_test.cpp
```diff
@@ -327,6 +327,9 @@ MT_TEST(DeviceInfo, basic) {
     if (queries[i] != CL_DEVICE_EXTENSIONS) {
       // Non-empty result for most supported queries.
       CHECK(size_ret > 0);
+    } else {
+      // Check buffer location is a supported extension
+      CHECK(strstr(str, "cl_intel_mem_alloc_buffer_location"));
     }
```
pcolberg, line 333: This check is unrelated to the `size_ret`, so I would suggest adding a separate section in the same style as the other query tests, i.e. `if (queries[i] == CL_DEVICE_EXTENSIONS) { ... }`.

This only tests for this particular extension, would you like to test for the other extensions too? Instead of hard-coding the extensions string, you could also compare for equality with `m_platform->extensions`, to avoid duplicating the list of extensions, e.g., as in [this test in the GPU runtime](https://github.com/intel/compute-runtime/blob/fc2f8cadfd450aa3071b9ef67b76edaadfd335bc/opencl/test/unit_test/device/device_caps_tests.cpp#L967). To ensure the test is robust (what if `m_platform->extensions` is corrupted for some reason?), you could factor out the extension string into a `#define` and use that to both initialize `m_platform->extensions` and to compare with in the unit test.

pcolberg (changes requested):
Thanks @sherry-yuan :slightly_smiling_face: 

Maybe title the commit `Add buffer location to supported extensions` since it is always supported?

test/acl_device_test.cpp
sherry-yuan: I have pushed few commits

the first on added the extension name and preliminary test
the second remove if statement check around the return size
the third added full checks for all extension.

Details are in the commit message.

sherry-yuan:
I think something is wrong with the windows machine, ninja got uninstalled?

https://github.com/intel/fpga-runtime-for-opencl/runs/5426146169?check_suite_focus=true

```
     | The term 'ninja' is not recognized as a name of a cmdlet, function, script file, or executable
     | program. Check the spelling of the name, or if a path was included, verify that the path is correct
     | and try again.
```

pcolberg:
> I think something is wrong with the windows machine, ninja got uninstalled?
> 
> https://github.com/intel/fpga-runtime-for-opencl/runs/5426146169?check_suite_focus=true
> 
> ```
>      | The term 'ninja' is not recognized as a name of a cmdlet, function, script file, or executable
>      | program. Check the spelling of the name, or if a path was included, verify that the path is correct
>      | and try again.
> ```

No worries, this looks like a temporary failure, maybe related to actions/cache. The [previous run](https://github.com/intel/fpga-runtime-for-opencl/runs/5425896867?check_suite_focus=true) failed on Windows 2022 with a similar error of `aoc` missing from `PATH`.

pcolberg (approved):
Thanks Sherry! I have revised the commit series slightly while keeping the contents the same.

When committing a change, think of it like a story where each commit contains a different aspect of your change. Rather than a chronological list, try to organise your commits into a logical series, where each commit is as self-contained as possible.

Taking this change as an example, you have three self-contained changes.
1. Checking for `size_ret > 0` in all cases. This is independent of the other changes, so you can make this change anywhere in the series; usually I prefer to get these out of the way first so the reviewer can focus on the actual change thereafter.
2. Factoring out and testing the extensions string. Since you want to proof that your desired change, adding the `buffer_location` extension, does not break the test, this commit should come before the next commit.
3. Now you can add the new extension. Not only have you proven that your change does not break the test, you have also shown that the test does not require any modification to test your new extension string, which was the point of factoring out the extension string.

The story-telling commit series makes sense when the maintainers value their code history, i.e., merge or rebase pull-request branches. Fine-grained, self-contained commits can be immensely valuable for later reading and comprehension, or easy cherry-picking to release branches, to give only two of many examples. On the other hand, when pull-request branches are squashed to a single commit, the effort is unfortunately in vain.

sherry-yuan:
Thanks @pcolberg for the guide!

In this case we would prefer a PR to contain multiple commits where each is a self-contained change rather than squashing all to one commit?

pcolberg:
> In this case we would prefer a PR to contain multiple commits where each is a self-contained change rather than squashing all to one commit?

Yes, because in this PR each commit introduces a new aspect of your change, multiple commits are justified.

On the other hand, consinder another example of one commit making a small change and the next commit applying `clang-format`. In this case, the two commits should be squashed, since the `clang-format` is nothing but noise to a reviewer, it does not add anything new to their understanding of the change.

If in doubt, I would always tend to more commits rather than fewer; commits can always be squashed if desired, but unsquashing a commit requires more work.

sherry-yuan:
I see, thanks!

PR #98, opened 2022-03-30
Backport non-code commits to 2022.2

pcolberg:
This backports small documentation and infrastructure changes to 2022.2. ~~Note how the list of workflow runs does **not** include Klocwork since build and source files are unchanged, as should be the case post code complete.~~ The list of workflow runs does include Klocwork due to [updating acl_auto.h for the 2022.2 release](https://github.com/intel/fpga-runtime-for-opencl/pull/98/commits/cf1412566db48b0bfd0eb0346b80f0371f55e158). :facepalm:

PR #97, opened 2022-03-30
Prepare 2022.2 release

pcolberg:
These commits are applied to the `main` branch first, then to the `2022.2` branch in https://github.com/intel/fpga-runtime-for-opencl/pull/98 using `git cherry-pick -x`. This updates `acl_auto.h` to `2022.2` first and separately to `2022.3`, so that the former commit may be cherry-picked to the `2022.2` branch, while the latter is applied to `main` **before** 2022.3 branching.

PR #99, opened 2022-03-31
Document releasing the runtime

pcolberg:
Document the release procedure followed in #97 and #98.


zibaiwan (approved):
Thanks Peter for documenting the process. It looks very clear to me.

PR #100, opened 2022-04-04
[Device Global] Autodiscovery change

sherry-yuan:
Autodiscovery help bridge the communication between backend and runtime.

The autodiscovery will contain the following information for a device global:
1. device global name
2. device global address
3. device global size

The autodiscovery will be in the following format:
`<number of device global> <number of field in each device global> [(<name>, <address>, <size>), (...), ...]`

The autodiscovery change came from https://github.com/intel/fpga-runtime-for-opencl/pull/65

### DIFF
--- a/include/acl.h
+++ b/include/acl.h
@@ -7,6 +7,7 @@
 #include <array>
 #include <assert.h>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include <CL/cl_ext.h>
@@ -478,6 +479,12 @@ typedef class acl_device_program_info_t *acl_device_program_info;
  */
 #define ACL_MEM_CAPABILITY_P2P (1 << 3)
 
+// Definition of device global.
+struct acl_device_global_mem_def_t {
+  uint32_t address;
+  uint32_t size;
+};
+
 // Part of acl_device_def_t where members are populated from the information
 // in the autodiscovery string. This will get updated every time the device
 // is programmed with a new device binary as the new binary would contain a
@@ -496,6 +503,10 @@ typedef struct acl_device_def_autodiscovery_t {
   std::array<acl_system_global_mem_def_t, ACL_MAX_GLOBAL_MEM> global_mem_defs;
 
   std::vector<acl_hostpipe_info_t> acl_hostpipe_info;
+
+  // Device global definition.
+  std::unordered_map<std::string, acl_device_global_mem_def_t>
+      device_global_mem_defs;
 } acl_device_def_autodiscovery_t;
 
 typedef struct acl_device_def_t {
--- a/src/acl_auto_configure.cpp
+++ b/src/acl_auto_configure.cpp
@@ -103,6 +103,25 @@ static bool read_uint_counters(const std::string &str,
 // Returns true if a valid integer was read or false if an error occurred.
 // pos is updated to the position immediately following the parsed word
 // even if an error occurs.
+static bool read_uint32_counters(const std::string &str,
+                                 std::string::size_type &pos, uint32_t &val,
+                                 std::vector<int> &counters) noexcept {
+  std::string result;
+  pos = read_word(str, pos, result);
+  decrement_section_counters(counters);
+  try {
+    val = static_cast<uint32_t>(std::stoul(result));
+  } catch (const std::exception &e) {
+    UNREFERENCED_PARAMETER(e);
+    return false;
+  }
+  return true;
+}
+
+// Reads the next word in str and converts it into an unsigned.
+// Returns true if a valid integer was read or false if an error occurred.
+// pos is updated to the position immediately following the parsed word
+// even if an error occurs.
 static bool read_int_counters(const std::string &str,
                               std::string::size_type &pos, int &val,
                               std::vector<int> &counters) noexcept {
@@ -491,6 +510,69 @@ bool acl_load_device_def_from_str(const std::string &config_str,
   if (result && counters.back() > 0) {
     result = read_uint_counters(config_str, curr_pos, kernel_arg_info_available,
                                 counters);
+  }
+
+  // Read device global information.
+  unsigned int num_device_global = 0;
+  if (result && counters.back() > 0) {
+    result =
+        read_uint_counters(config_str, curr_pos, num_device_global, counters);
+
+    // read total number of fields in device global
+    unsigned int total_fields_device_global = 0;
+    if (result) {
+      result = read_uint_counters(config_str, curr_pos,
+                                  total_fields_device_global, counters);
+    }
+
+    for (auto i = 0U; result && (i < num_device_global);
+         i++) { // device_global_memories
+      counters.emplace_back(total_fields_device_global);
+
+      // read device global name
+      std::string device_global_name;
+      if (result && counters.back() > 0) {
+        result = read_string_counters(config_str, curr_pos, device_global_name,
+                                      counters);
+      }
+
+      // read device global address
+      uint32_t dev_global_addr = 0; // Default
+      if (result && counters.back() > 0) {
+        result = read_uint32_counters(config_str, curr_pos, dev_global_addr,
+                                      counters);
+      }
+      // read device global address size
+      uint32_t dev_global_size = 0; // Default
+      if (result && counters.back() > 0) {
+        result = read_uint32_counters(config_str, curr_pos, dev_global_size,
+                                      counters);
+      }
+
+      acl_device_global_mem_def_t dev_global_def = {dev_global_addr,
+                                                    dev_global_size};
+      bool ok = devdef.device_global_mem_defs
+                    .insert({device_global_name, dev_global_def})
+                    .second;
+      if (!ok) {
+        // Device global name already exist in map, but it should have been
+        // unique.
+        err_ss << "Device global name should be unique. " << device_global_name
+               << " is repeated.\n";
+        result = false;
+      }
+
+      // forward compatibility: bypassing remaining fields at the end of device
+      // global memory
+      while (result && counters.size() > 0 &&
+             counters.back() > 0) { // total_fields_device_global>0
+        std::string tmp;
+        result =
+            result && read_string_counters(config_str, curr_pos, tmp, counters);
+        check_section_counters(counters);
+      }
+      counters.pop_back(); // removing total_fields_device_global
+    }                      // device_global_memories
   }
 
   // forward compatibility: bypassing remaining fields at the end of device
--- a/test/acl_auto_configure_test.cpp
+++ b/test/acl_auto_configure_test.cpp
@@ -36,6 +36,7 @@ TEST(auto_configure, simple) {
 #define VERSIONIDSTRINGIFY(x) #x
 #define VERSIONIDTOSTR(x) VERSIONIDSTRINGIFY(x)
 #define DEVICE_FIELDS " 23"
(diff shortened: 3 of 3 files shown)


include/acl.h
```diff
@@ -478,6 +479,13 @@ typedef class acl_device_program_info_t *acl_device_program_info;
  */
 #define ACL_MEM_CAPABILITY_P2P (1 << 3)
 
+// Definition of device global.
+typedef struct acl_device_global_mem_def_t {
+  std::string name;
+  unsigned int address;
+  unsigned int size;
```
pcolberg, line 486: Could you confirm this is indeed the desired integer type? `unsigned int` basically says that you want an unsigned integer that is *at least* 32 bits (on the architectures supported by the runtime) but might be larger. `uint32_t` would be more appropriate for *exactly* 32 bits.

include/acl.h
```diff
@@ -478,6 +479,13 @@ typedef class acl_device_program_info_t *acl_device_program_info;
  */
 #define ACL_MEM_CAPABILITY_P2P (1 << 3)
 
+// Definition of device global.
+typedef struct acl_device_global_mem_def_t {
```
pcolberg, line 483: `typedef` is not needed in C++ to directly refer to the type without `struct`.

include/acl.h
```diff
@@ -478,6 +479,13 @@ typedef class acl_device_program_info_t *acl_device_program_info;
...
+  std::string name;
```
pcolberg, line 484: Is the duplicate `name` needed, since it is already used as the unique key?

src/acl_auto_configure.cpp
```diff
@@ -493,6 +493,61 @@ bool acl_load_device_def_from_str(const std::string &config_str,
...
 
+  // Read device global information.
+  unsigned int num_device_global = 0;
+  if (result && counters.back() > 0) {
+    result =
+        read_uint_counters(config_str, curr_pos, num_device_global, counters);
+    devdef.num_device_global = num_device_global;
+
+    // read total number of fields in device global
+    int total_fields_device_global = 0;
```
pcolberg, line 504: Why is the integer signed?

src/acl_auto_configure.cpp
```diff
@@ -493,6 +493,61 @@ bool acl_load_device_def_from_str(const std::string &config_str,
...
+    for (auto i = 0U; result && (i < num_device_global);
+         i++) { // device_global_memories
+      counters.emplace_back(total_fields_device_global);
+
+      // read device global name
+      std::string device_global_name;
+      if (result && counters.back() > 0) {
+        result = read_string_counters(config_str, curr_pos, device_global_name,
+                                      counters);
+      }
```
pcolberg, line 519: What happens if `result` evaluates to `false` and `device_global_name` is uninitialized, i.e. the empty string?

src/acl_auto_configure.cpp
```diff
@@ -493,6 +493,61 @@ bool acl_load_device_def_from_str(const std::string &config_str,
...
+      // read device global address size
+      unsigned int dev_global_size = 0; // Default
+      if (result && counters.back() > 0) {
+        result =
+            read_uint_counters(config_str, curr_pos, dev_global_size, counters);
+      }
+
+      acl_device_global_mem_def_t dev_global_def = {
+          device_global_name, dev_global_addr, dev_global_size};
+      devdef.device_global_mem_defs[device_global_name] = dev_global_def;
```
pcolberg, line 536: If the key does not exist yet, this will insert a default-constructed value and then overwrite with the given value. If the key does exist, this will overwrite the existing value. You can use [`std::unordered_map::insert`](https://en.cppreference.com/w/cpp/container/unordered_map/insert) to avoid the default construction when the key does not exist, and to signal an error when the key does exist.

test/acl_auto_configure_test.cpp
```diff
@@ -261,6 +269,27 @@ TEST(auto_configure, simple) {
               (int)m_device_def.autodiscovery_def.accel[0].max_work_group_size);
   CHECK_EQUAL(1, (int)m_device_def.autodiscovery_def.accel[0].is_sycl_compile);
 
+  // Checks for device global entry.
+  CHECK_EQUAL(2, (int)m_device_def.autodiscovery_def.num_device_global);
```
pcolberg, line 273: Granted the above is not better, but do it the other way around: `2u` instead of casting the value under test.

The rationale is that you know `2u` is a valid unsigned integer, but you don't know what the range of the value under test is in case of a failure. You want to avoid casting a potentially unknown value to a different type, which in the worst case might pass your test despite a wrong result.

test/acl_auto_configure_test.cpp
```diff
@@ -261,6 +269,27 @@ TEST(auto_configure, simple) {
...
+  CHECK(m_device_def.autodiscovery_def.device_global_mem_defs.find(
+            "kernel15_dev_global") !=
+        m_device_def.autodiscovery_def.device_global_mem_defs.end());
+  CHECK(m_device_def.autodiscovery_def.device_global_mem_defs.find(
+            "kernel15_dev_global2") !=
+        m_device_def.autodiscovery_def.device_global_mem_defs.end());
+  CHECK_EQUAL(4096, m_device_def.autodiscovery_def
+                        .device_global_mem_defs["kernel15_dev_global"]
```
pcolberg, line 281: Instead of doing a write access using [`[]`](https://en.cppreference.com/w/cpp/container/unordered_map/operator_at), store the return values of [`find`](https://en.cppreference.com/w/cpp/container/unordered_map/find), then use them here to access the members. (This would not compile if `m_device_def` were `const`, hence constify all the things :slightly_smiling_face:).

test/acl_auto_configure_test.cpp
pcolberg: Then again, why is this needed here and not below, where `size` and `address` are also unsigned?

pcolberg (changes requested):
Thanks @sherry-yuan for your great work on `device_global` :slightly_smiling_face:

src/acl_auto_configure.cpp
sherry-yuan: if any one of the result is false, it will get propagated until the end of the function (i.e once its false, it will be false for the lifetime of the function call). and all the acceleration info is empty then output error.

include/acl.h
sherry-yuan: size could technically be uint32_t, I applied the change.

test/acl_auto_configure_test.cpp
sherry-yuan: Thanks for all the suggestions,  changes are made accordingly in https://github.com/intel/fpga-runtime-for-opencl/pull/100/commits/b2bdd1201252c7e059c1505345d94fcadecb6aef

include/acl.h
pcolberg: Why is `size` 32 bit and not 64 bit? Does this also apply to `address`?

test/acl_auto_configure_test.cpp
```diff
@@ -261,6 +269,27 @@ TEST(auto_configure, simple) {
               (int)m_device_def.autodiscovery_def.accel[0].max_work_group_size);
   CHECK_EQUAL(1, (int)m_device_def.autodiscovery_def.accel[0].is_sycl_compile);
 
+  // Checks for device global entry.
+  CHECK_EQUAL(2, m_device_def.autodiscovery_def.num_device_global);
+  CHECK(m_device_def.autodiscovery_def.device_global_mem_defs.find(
+            "kernel15_dev_global") !=
+        m_device_def.autodiscovery_def.device_global_mem_defs.end());
```
pcolberg, line 276: You can store the iterator:
```suggestion
  const auto kernel15_dev_global = m_device_def.autodiscovery_def.device_global_mem_defs.find(
            "kernel15_dev_global");
  CHECK(kernel15_dev_global !=
        m_device_def.autodiscovery_def.device_global_mem_defs.end());
```

test/acl_auto_configure_test.cpp
```diff
@@ -261,6 +269,27 @@ TEST(auto_configure, simple) {
...
+  CHECK(m_device_def.autodiscovery_def.device_global_mem_defs.find(
+            "kernel15_dev_global2") !=
+        m_device_def.autodiscovery_def.device_global_mem_defs.end());
+  CHECK_EQUAL(4096, m_device_def.autodiscovery_def.device_global_mem_defs
+                        .find("kernel15_dev_global")
+                        ->second.address);
```
pcolberg, line 282: Then use the iterator for further checks:
```suggestion
  CHECK_EQUAL(4096, kernel15_dev_global
                        ->second.address);
```

pcolberg (changes requested):
Thanks @sherry-yuan, looks good! I am still unsure about the integer types which you should be able to answer by copying what the compiler does.

include/acl.h
sherry-yuan: On the frontend, they defined the size as 32 bit. https://github.com/intel/llvm/blob/88123c1c42e71aa3667d24558ed9c9baca39467a/llvm/tools/sycl-post-link/DeviceGlobals.cpp#L61

sherry-yuan: I think address is going to be converted from some pointer value, Its safer to keep it unsigned int

test/acl_auto_configure_test.cpp
sherry-yuan: Thanks for the suggestions, I have applied the changes and squashed the commits.

include/acl.h
```diff
@@ -496,6 +503,11 @@ typedef struct acl_device_def_autodiscovery_t {
   std::array<acl_system_global_mem_def_t, ACL_MAX_GLOBAL_MEM> global_mem_defs;
 
   std::vector<acl_hostpipe_info_t> acl_hostpipe_info;
+
+  // Device global definition.
+  unsigned int num_device_global;
```
pcolberg, line 508: Apologies for missing this the first time: Can `num_device_global` be removed, in favour of using `device_global_mem_defs.size()` which is `O(1)` in time? The idea is to make as much use as possible of the self-describing properties of C++ containers.

include/acl.h
pcolberg: Thanks for confirming `size`. As for `address`, if it stores a pointer value you might want to use `uintptr_t`. But I would recommend you do try to find the corresponding type in the compiler and use that here.

pcolberg (changes requested):
Thanks @sherry-yuan for your patience, almost done :sweat_smile:

include/acl.h
sherry-yuan: I checked the IR generated for device global, if I am interpreting it correctly, the type should be uint32 for address as well: https://github.com/intel/llvm/blob/88123c1c42e71aa3667d24558ed9c9baca39467a/llvm/test/tools/sycl-post-link/device-globals/test_no_property_set_header_for_an_empty_set.ll#L11

include/acl.h
sherry-yuan: Thanks! Done.

pcolberg: Perfect. I think you forgot to remove the member. 
```suggestion
```

pcolberg (changes requested):

include/acl.h
sherry-yuan: Thanks for removing it. Sorry about that 🤦

pcolberg (approved):

PR #104, opened 2022-04-14
Initialize max number of global memory definition for simulator

sherry-yuan:
Simulator does not have any global memory interface information until the actuall aocx is loaded.
(Note this is only a problem for simulator not hardware run, in hardware run, we can communicate with BSP to query memory interface information)
Prior to loading aocx it uses predefined autodiscovery [1] to initialize its global memory interface, which has only 1 global memory
In the sycl runtime flow today, the USM device allocation call happens before aocx is loaded.
The aocx is loaded when clCreateProgram is called, which typically happen on first kernel launch in sycl runtime.
The USM device allocation on mutli global memory system will fail because there are in total 1 global memory as defined in [1] but the user is requesting more than 1 device global memory.
User could go around this issue by launching a sacrificial kernel that uses shared allocation as kernel argument. This will setup the correct global memory interface in runtime.
This change eliminate the need to run a sacrificial kernel.
However there are a few downside:
1. The address range/size may not be exactly the same as the one that is in aocx, but this is not too large of a problem because runtime first fit allocation algorithm will fill the lowest address range first. Unless user requested more than what is availble.
2. it potentially occupied more space than required
3. will not error out when user requested a non-existing device global memory because we are using ACL_MAX_GLOBAL_MEM for num_global_mem_systems

[1] https://github.com/intel/fpga-runtime-for-opencl/blob/950f21dd079dfd55a473ba4122a4a9dca450e36f/include/acl_shipped_board_cfgs.h#L7

### DIFF
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -715,6 +715,29 @@ int acl_kernel_if_init(acl_kernel_if *kern, acl_bsp_io bsp_io,
     auto parse_result = acl_load_device_def_from_str(
         std::string(acl_shipped_board_cfgs[1]),
         sysdef->device[0].autodiscovery_def, err_msg);
+    // Fill in definition for all device global memory
+    // Simulator does not have any global memory interface information until the
+    // actual aocx is loaded. (Note this is only a problem for simulator not
+    // hardware run, in hardware run, we can communicate with BSP to query
+    // memory interface information). In the flow today, the USM device
+    // allocation call happens before aocx is loaded. The aocx is loaded when
+    // clCreateProgram is called, which typically happen on first kernel launch
+    // in sycl runtime. In order to prevent the USM device allocation from
+    // failing on  mutli global memory system, initialize as much global memory
+    // system as possible for simulation flow. However there are a few downside:
+    // 1. The address range/size may not be exactly the same as the one that is
+    // in aocx, but this is not too large of a problem because runtime first fit
+    // allocation algorithm will fill the lowest address range first. Unless
+    // user requested more than what is availble.
+    // 2. it potentially occupied more space than required
+    // 3. will not error out when user requested a non-existing device global
+    // memory because we are using ACL_MAX_GLOBAL_MEM for num_global_mem_systems
+    sysdef->device[0].autodiscovery_def.num_global_mem_systems =
+        ACL_MAX_GLOBAL_MEM;
+    for (int i = 0; i < ACL_MAX_GLOBAL_MEM; i++) {
+      sysdef->device[0].autodiscovery_def.global_mem_defs[i] =
+          sysdef->device[0].autodiscovery_def.global_mem_defs[0];
+    }
     if (parse_result)
       sysdef->num_devices = 1;
     // Override the device name to the simulator.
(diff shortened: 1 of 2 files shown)


pcolberg:
Looks like git checkout is broken. Maybe GitHub changed their container setup?
https://github.com/intel/fpga-runtime-for-opencl/runs/6029451177?check_suite_focus=true
```
2022-04-14T19:08:35.1638886Z ##[group]Run sudo chown -R build:build .
2022-04-14T19:08:35.1639231Z [36;1msudo chown -R build:build .[0m
2022-04-14T19:08:35.1643269Z shell: sh -e {0}
2022-04-14T19:08:35.1643503Z ##[endgroup]
2022-04-14T19:08:35.3237562Z ##[group]Run actions/checkout@v2
2022-04-14T19:08:35.3237813Z with:
2022-04-14T19:08:35.3238075Z   repository: intel/fpga-runtime-for-opencl
2022-04-14T19:08:35.3238576Z   token: ***
2022-04-14T19:08:35.3238765Z   ssh-strict: true
2022-04-14T19:08:35.3238996Z   persist-credentials: true
2022-04-14T19:08:35.3239223Z   clean: true
2022-04-14T19:08:35.3239407Z   fetch-depth: 1
2022-04-14T19:08:35.3239606Z   lfs: false
2022-04-14T19:08:35.3239800Z   submodules: false
2022-04-14T19:08:35.3239993Z ##[endgroup]
2022-04-14T19:08:35.3334732Z ##[command]/usr/bin/docker exec  5b270e3f98b5ba860472430574ecd30e726cd60a8663a6981828793fee29b786 sh -c "cat /etc/*release | grep ^ID"
2022-04-14T19:08:35.8300296Z Syncing repository: intel/fpga-runtime-for-opencl
2022-04-14T19:08:35.8301911Z ##[group]Getting Git version info
2022-04-14T19:08:35.8302459Z Working directory is '/__w/fpga-runtime-for-opencl/fpga-runtime-for-opencl'
2022-04-14T19:08:35.8303000Z [command]/usr/bin/git version
2022-04-14T19:08:35.8303819Z git version 2.27.0
2022-04-14T19:08:35.8304529Z ##[endgroup]
2022-04-14T19:08:35.8357485Z ##[error]EACCES: permission denied, mkdir '/__w/_temp/dfdaa6c0-6817-4c98-9c22-84b0ea9c776a'
```

sherry-yuan:
No regression was found in external&internal tests. This is ready to merge.

pcolberg (changes requested):
Thanks @sherry-yuan, looks good!

1. The commit message contains very useful background. What do you think about moving the commit message to a comment in the source code?

2. @bsyrowik committed a similar change in d9df7a9ed68d1343342666a0466d154561599a1a, although it sets 128 memory systems instead of [ACL_MAX_GLOBAL_MEM](https://github.com/intel/fpga-runtime-for-opencl/blob/dfdd0d85eee8519e973b5d12d0297599fc4fb14d/include/acl.h#L50) == 32. With your change that commit is no longer needed, correct? Could you add a `git revert d9df7a9e` to your pull request? Is there any downside to setting the default 32 instead of the non-default 128 memory systems?

sherry-yuan:
> 1. What do you think about moving the commit message to a comment in the source code?

Great idea, I have moved the commit message to line comment with slight modification.

> With your change that commit is no longer needed, correct?

Right its no longer needed

> Is there any downside to setting the default 32 instead of the non-default 128 memory systems?

I do not think so, as far as I know, all the device so far support up to 32 global memory, therefore there is no point for simulation to support more than 32 global memory.

pcolberg (approved):

PR #105, opened 2022-04-15
Override uid of default user in container to align with GitHub-hosted runners

pcolberg:
This resolves a recent regression with GitHub-hosted runners where
actions/checkout@v2 fails to create a directory under /__w/_temp/
    
https://github.com/intel/fpga-runtime-for-opencl/pull/104#issuecomment-1099590565
    
https://github.com/actions/checkout/issues/47


zibaiwan (approved):
Thanks Peter! LGTM!

PR #107, opened 2022-04-21
Remove unneeded extern "C" from acl_auto_configure.h

pcolberg:
While the use of extern "C" is technically valid and results in a
C-compatible symbol name that does not include return and argument
types, the function takes a std::string which is unavailable in C.


zibaiwan (approved):
LGTM! Thanks Peter!

PR #108, opened 2022-04-22
Factor out functions for auto-discovery parsing

pcolberg:
This factors out parsing of auto-discovery string sections into separate functions to improve the readability and maintenance of `acl_load_device_def_from_str()`. Care was taken not to modify the functional behaviour, by carrying out the refactoring twice and ensuring that the results are identical.

A choice had to be made between factoring out parsing of individual section entries, e.g., a single kernel argument, versus factoring out entire loops, e.g., all kernel arguments. The latter was chosen since the data structures vary (`vector` versus `unordered_map`), and, importantly, the number of fields vary between variable and fixed counts. For [variable count](https://github.com/intel/fpga-runtime-for-opencl/blob/a446b30c59e3c0b45eeeb0a1897dbbddc9bd9e9e/src/acl_auto_configure.cpp#L695-L700), the number of fields is read for each entry; whereas for [fixed count](https://github.com/intel/fpga-runtime-for-opencl/blob/a446b30c59e3c0b45eeeb0a1897dbbddc9bd9e9e/src/acl_auto_configure.cpp#L454-L462), the number of fields is read once for all entries. By factoring out the entire loop, the field count parsing is contained within the function in both cases.

### DIFF
--- a/src/acl_auto_configure.cpp
+++ b/src/acl_auto_configure.cpp
@@ -193,6 +193,185 @@ static int read_string_counters(const std::string &str,
   return result != "";
 }
 
+static bool
+read_global_mem_defs(const std::string &config_str,
+                     std::string::size_type &curr_pos,
+                     unsigned int &num_global_mem_systems,
+                     std::array<acl_system_global_mem_def_t, ACL_MAX_GLOBAL_MEM>
+                         &global_mem_defs,
+                     std::vector<int> &counters) noexcept {
+  bool result = read_uint_counters(config_str, curr_pos, num_global_mem_systems,
+                                   counters);
+
+  for (auto i = 0U; result && (i < num_global_mem_systems); i++) {
+    std::string gmem_name;
+    // read total number of fields in global_memories
+    int total_fields_global_memories = 0;
+    result = read_int_counters(config_str, curr_pos,
+                               total_fields_global_memories, counters);
+    counters.emplace_back(total_fields_global_memories);
+
+    // read global memory name
+    if (result) {
+      result = read_string_counters(config_str, curr_pos, gmem_name, counters);
+    }
+
+    // read global memory type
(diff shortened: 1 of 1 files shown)


src/acl_auto_configure.cpp
```diff
@@ -193,6 +193,125 @@ static int read_string_counters(const std::string &str,
...
+    result =
+        result && read_uint_counters(config_str, curr_pos, alignment, counters);
+  }
+
+  std::string buffer_location = "";
+  if (result) {
+    result = result && read_uint_counters(config_str, curr_pos,
+                                          num_buffer_locations, counters);
+    for (auto k = 0U; result && (k < num_buffer_locations); k++) {
+      if (k /* FIXME j */ > 0) {
```
pcolberg, line 228: This previously checked the argument index `j`, but likely intended to check the attribute index `k`.

src/acl_auto_configure.cpp
```diff
@@ -193,6 +193,125 @@ static int read_string_counters(const std::string &str,
...
+        buffer_location = buffer_location + " ";
+      }
+
+      result = result && read_string_counters(config_str, curr_pos,
+                                              buffer_location, counters);
```
pcolberg, line 233: This actually overwrites `buffer_location`, instead of appending as likely intended.

src/acl_auto_configure.cpp
pcolberg: This has been resolved in #108

src/acl_auto_configure.cpp
pcolberg: This has been resolved in #108

sherry-yuan (approved):
Thanks Peter, the restructuring looks good to me.

(merge after tests passes tho)

PR #103, opened 2022-04-13
Support streaming kernel arguments and streaming kernel control

pcolberg:
This introduces a backwards-incompatible change in the HAL MMD API between runtime and simulator. The first commit defines the streaming kernel argument interface, while subsequent commits will contain the implementation.

@anandhv @bsyrowik @zibaiwan @sherry-yuan @intel-jisheng1

PR #109, opened 2022-04-22
Print warning when kernel argument has multiple buffer_location attributes

pcolberg:
The auto-discovery string parsing of per kernel argument buffer_location
intended to concatenate multiple attributes into a space-separated string,
but in fact only retains the value of the last buffer_location attribute
since read_string_counters() does not append but assign the string value.

Since multiple buffer_location attributes are not supported, print a
warning message, and select the last buffer_location attribute such that
the functional behaviour of buffer_location parsing remains unchanged.

This is a prerequisite of #108 and #103, since the current implementation
incorrectly checks argument index `j` instead of attribute index `k` which
results in a build failure when factoring out kernel argument parsing into
a separate function.

### DIFF
--- a/src/acl_auto_configure.cpp
+++ b/src/acl_auto_configure.cpp
@@ -685,7 +685,6 @@ bool acl_load_device_def_from_str(const std::string &config_str,
         auto addr_space_type = 0U;
         auto category = 0U;
         auto size = 0U;
-        auto num_buffer_locations = 0U;
         int total_fields_arguments = 0;
         if (result) {
           result =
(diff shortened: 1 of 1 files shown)


src/acl_auto_configure.cpp
```diff
@@ -706,13 +705,16 @@ bool acl_load_device_def_from_str(const std::string &config_str,
...
-          for (auto k = 0U; result && (k < num_buffer_locations); k++) {
-            if (j > 0) {
-              buffer_location = buffer_location + " ";
-            }
-
+          if (result && num_buffer_locations > 1) {
+            std::cerr
+                << "WARNING: kernel argument has multiple buffer_location "
+                   "attributes which is not supported.\nSelecting last "
+                   "buffer_location attribute.\n";
```
sherry-yuan, line 715: I think it might be nicer to let user know which is the last one, I am not sure if the order of the buffer location is the same as how they passed it into the property list.

e.g. a format string may look like: `("Warning: kernel argument has multiple buffer_location attributes which is not supported. Selecting %s as buffer location for kernel argument %d", buffer_location, j)`

pcolberg: Thanks Sherry, that's a good suggestion. The attribute value is easy to print, however printing the argument index is slightly trickier since `j` is no longer available when this code is factored out (which uncovered that this code was buggy in the first place). A proper solution would be to have hierarchical error handling, i.e. the outer function prints "error while parsing kernel argument %d" and the inner function prints "selecting %s as buffer location". However, since this code is currently not unit-tested, I would like to keep refactoring to a minimum and only print the attribute value for now.

pcolberg: I have updated the message to include `buffer_location`. As mentioned above, including the kernel argument index would be useful to the user, but it will have to wait for later refactoring, after increasing code coverage.

zibaiwan (approved):
Thanks Peter and Sherry! LGTM!

sherry-yuan (approved):
Thanks Peter :)

PR #110, opened 2022-04-22
Make auto-discovery error message handling more failsafe

pcolberg:
Currently it is too easy to fill the stringstream err_ss with an error
message but forget to assign the contents to the string err_str.
    
Reduce scope of stringstream err_ss to where it is used to make such
mistakes more obvious. In C++20, this may be replaced with std::format().

This is a prerequisite of #108 and #103.


zibaiwan (approved):
Thanks Peter! LGTM!

PR #114, opened 2022-05-04
l_copy_and_adjust_arguments_for_device: reduce scope of device-side buffer increment

pcolberg:
This ensures the buffer increment is initialized to zero for every
kernel argument, not only the first one. This has no functional impact
since all kernel argument types use a non-default buffer increment.

This is a prerequisite for https://github.com/intel/fpga-runtime-for-opencl/pull/103.


zibaiwan (approved):
Looks good! Thanks Peter!

PR #115, opened 2022-05-06
invocation_wrapper: support non-POD types using placement new

pcolberg:
Each context maintains a pool of manually allocated invocation_wrapper
objects, which so far contained only POD types. To support the use of
non-POD types, e.g., std::vector, explicitly invoke the constructor
using placement new after malloc, and the destructor before free.

https://en.cppreference.com/w/cpp/language/new#Placement_new
https://github.com/intel/fpga-runtime-for-opencl/pull/103#discussion_r866418903


zibaiwan (approved):
Looks good! Thanks Peter!

PR #139, opened 2022-05-25
Explicitly clear the CSR_done bit

shuoniu-intel:
Also, update the name of some CSR bit macros to match the terminology in the compiler KernelCRA.

Tested locally in qshell2, passed L2.


src/acl_kernel_if.cpp
```diff
@@ -44,7 +44,8 @@ acl_process_autorun_profiler_scan_chain(unsigned int physical_device_id,
 // to verify that the hardware and HAL have an identical view of the CSR
 // address map.
 #define CSR_VERSION_ID_181 (3)
-#define CSR_VERSION_ID (4)
+#define CSR_VERSION_ID_20222 (4)
```
pcolberg, line 47: Since the versioning scheme changed from FPGA SDK to oneAPI anyway, how about using `_` inplace of `.`?
```suggestion
#define CSR_VERSION_ID_2022_2 (4)
```

src/acl_kernel_if.cpp
```diff
@@ -90,8 +91,8 @@ acl_process_autorun_profiler_scan_chain(unsigned int physical_device_id,
 #define KERNEL_CSR_LMEM_INVALID_BANK 11
 #define KERNEL_CSR_LSU_ACTIVE 12
 #define KERNEL_CSR_WR_ACTIVE 13
-#define KERNEL_CSR_VALID_IN 14
-#define KERNEL_CSR_STARTED 15
+#define KERNEL_CSR_BUSY 14
+#define KERNEL_CSR_RUNNING 15
```
pcolberg, line 95: Could you commit this as a separate change and motivate the renaming in the commit message?

Each runtime commit should ideally be atomic, i.e., a single, self-consistent change.

src/acl_kernel_if.cpp
```diff
@@ -1470,8 +1472,10 @@ void acl_kernel_if_update_status(acl_kernel_if *kern) {
 
       // Just clear the "done" bit.  The "go" bit should already have been
       // cleared, but this is harmless anyway.
-      // Since csr version 19, done bit is cleared when finish counter is read
-      if (kern->csr_version == CSR_VERSION_ID_181) {
+      // Since csr version 19, done bit is cleared when finish counter is read.
+      // Since csr version 2022.3, done bit needs to be cleared explicitly.
+      if (kern->csr_version == CSR_VERSION_ID_181 ||
+          kern->csr_version == CSR_VERSION_ID) {
```
pcolberg, line 1478: The above will break when the CSR version is bumped once more.
```suggestion
      // Since csr version 19, done bit is cleared when finish counter is read.
      // Since csr version 2022.3, done bit needs to be cleared explicitly.
      if (kern->csr_version == CSR_VERSION_ID_181 ||
          kern->csr_version >= CSR_VERSION_ID_2022_3) {
```

src/acl_kernel_if.cpp
```diff
@@ -1138,6 +1139,7 @@ void acl_kernel_if_launch_kernel_on_custom_sof(
   // in the kern->csr_version field.  The value is cached after PLL init to
   // avoid reading back on every kernel launch, which would add overhead.
   if (kern->csr_version != CSR_VERSION_ID &&
+      kern->csr_version != CSR_VERSION_ID_20222 &&
       kern->csr_version != CSR_VERSION_ID_181) {
```
pcolberg, line 1143: How about using a range for future compatibility?
```suggestion
  if (!(kern->csr_version >= CSR_VERSION_ID_181 &&
        kern->csr_version <= CSR_VERSION_ID)) {
```

src/acl_kernel_if.cpp
```diff
@@ -44,7 +44,8 @@ acl_process_autorun_profiler_scan_chain(unsigned int physical_device_id,
 // to verify that the hardware and HAL have an identical view of the CSR
 // address map.
 #define CSR_VERSION_ID_181 (3)
-#define CSR_VERSION_ID (4)
+#define CSR_VERSION_ID_20222 (4)
+#define CSR_VERSION_ID (5)
```
pcolberg, line 48: We could introduce an alias for the latest version:
```suggestion
#define CSR_VERSION_ID_2022_3 (5)
#define CSR_VERSION_ID CSR_VERSION_ID_2022_3
```
The explicitly versioned constant may then be used below.

pcolberg (changes requested):
Thanks @shuoniu-intel, looks good! I have made some minor suggestions to ease future bumps of the CSR version, as well as breaking up the rename of CSR bits into a separate commit.

src/acl_kernel_if.cpp
shuoniu-intel: Sure. I'll remove this renaming change from this patch.

pcolberg: Thanks @shuoniu-intel! You can still have the renaming in this PR, just a separate commit. We use the rebase-merge strategy for this repository, so your multiple commits will be preserved as is on merge into `main`.

shuoniu-intel: Cool! Good to know that commits in a PR will not be combined after merging in this repository.

src/acl_kernel_if.cpp
```diff
@@ -44,7 +44,9 @@ acl_process_autorun_profiler_scan_chain(unsigned int physical_device_id,
 // to verify that the hardware and HAL have an identical view of the CSR
 // address map.
 #define CSR_VERSION_ID_181 (3)
-#define CSR_VERSION_ID (4)
+#define CSR_VERSION_ID_2022_2 (4)
```
pcolberg, line 47: On second thought this should be named after the first release in which it was introduced, which is [19.1](https://www.intel.com/content/www/us/en/software-kit/664561/intel-fpga-sdk-for-opencl-pro-edition-software-version-19-1.html).
```suggestion
#define CSR_VERSION_ID_19_1 (4)
```

pcolberg (changes requested):

pcolberg:
Thanks @shuoniu-intel! I will merge once the integration testing has passed.

pcolberg (approved):

PR #141, opened 2022-06-01
Add new buffer location memory property to aocl_mmd.h

bsyrowik:


pcolberg:
Thanks Bain! Could you provide a brief description motivating the change?

include/MMD/aocl_mmd.h
```diff
@@ -484,7 +484,13 @@ typedef enum {
...
    *  AOCL_MMD_GLOBAL_MEMORY_INTERFACE.
    */
-  AOCL_MMD_MEM_PROPERTIES_MEMORY_BANK
+  AOCL_MMD_MEM_PROPERTIES_MEMORY_BANK,
+  /**
+   *  Specifies the buffer location used for USM API calls.  The buffer location
+   *  is an integer index that corresponds to a global memory location in the
+   *  board_spec.xml file.  Similar to AOCL_MMD_MEM_PROPERTIES_GLOBAL_MEMORY.
+   */
+  AOCL_MMD_MEM_PROPERTIES_BUFFER_LOCATION
```
pcolberg, line 493: I believe C++11 allows trailing comma now, which is nice for minimizing changes and conflicts :slightly_smiling_face: 
```suggestion
  AOCL_MMD_MEM_PROPERTIES_BUFFER_LOCATION,
```

pcolberg (dismissed):

pcolberg:
I suggest we add the runtime implementation to this PR that demonstrate how the new API would work. The idea is to pass the newly received `buffer_location` property in [`clSharedMemAllocINTEL()`](https://github.com/intel/fpga-runtime-for-opencl/blob/c7af3b57be92a0a7f26ecfb3807516ee48620157/src/acl_usm.cpp#L302-L457) and [`clHostMemAllocINTEL()`](https://github.com/intel/fpga-runtime-for-opencl/blob/c7af3b57be92a0a7f26ecfb3807516ee48620157/src/acl_usm.cpp#L45-L188) through to the MMD using this new attribute, correct? Besides that, is any additional logic needed in these runtime functions?

include/MMD/aocl_mmd.h
bsyrowik: Oh really!?  That is fantastic!  Is there a guarantee that users of this header will be compiling with C++11 or greater?

src/acl_usm.cpp
```diff
@@ -380,6 +394,7 @@ clSharedMemAllocINTEL(cl_context context, cl_device_id device,
   // The end of the properties list is specified with a zero.
   cl_mem_alloc_flags_intel alloc_flags = 0;
   std::unordered_set<unsigned long long> seen_flags;
+  cl_int mem_id = -1;
```
pcolberg, line 397: Would `std::optional<cl_int>` work here? To prevent future mistakes in using an uninitialized value.

src/acl_usm.cpp
```diff
@@ -396,6 +411,13 @@ clSharedMemAllocINTEL(cl_context context, cl_device_id device,
       }
       alloc_flags = *(properties + 1);
     } break;
+    case CL_MEM_ALLOC_BUFFER_LOCATION_INTEL: {
+      if (seen_flags.insert(CL_MEM_ALLOC_BUFFER_LOCATION_INTEL).second == false) {
+        UNLOCK_BAIL_INFO(CL_INVALID_PROPERTY, context,
+                         "Property specified multiple times");
+      }
+      mem_id = (cl_int) * (properties + 1);
```
pcolberg, line 419: Should this validate the property value passed in from the (OpenCL) user and return an error?

bsyrowik: I was wondering about this as well.  The issue is that we may not have loaded the appropriate boardspec/autodiscovery string, so we actually have no way of knowing which buffer locations are valid.  I'm adding some functionality to the sim MMD to preload the boardspec, so I can do checking there.  I think in general it is reasonable to push this responsibility on to the MMD.  I'll make a note about that in `aocl_mmd.h`.

bsyrowik: Lower bound of 0 is a good point.  I changed it to cl_uint, so shouldn't need to worry about that any more.  As for the upper limit, I don't think there is any way to enforce this.  There may be a use-case for having arbitrary buffer locations?  Not sure.

pcolberg: > I changed it to cl_uint, so shouldn't need to worry about that any more.

The [SYCL extension specifies a signed integer](https://github.com/intel/llvm/blob/db037ca70e1278bf5ee9ccb5c91d9e10ce7fb418/sycl/doc/extensions/proposed/sycl_ext_intel_runtime_buffer_location.asciidoc?plain=1#L146-L154) though, so we would either need to match that in the runtime, or change the SYCL extension which is still experimental.

@tiwaria1 What is the reason for using a signed integer for the buffer location property in the SYCL API?

src/acl_usm.cpp
```diff
@@ -138,6 +144,13 @@ CL_API_ENTRY void *CL_API_CALL clHostMemAllocINTEL(
   }
 
   if (acl_get_hal()->host_alloc) {
+    std::vector<mem_properties_t> mmd_properties;
+    if (mem_id) {
+      mmd_properties.push_back(AOCL_MMD_MEM_PROPERTIES_BUFFER_LOCATION);
+      mmd_properties.push_back(mem_id.value());
```
pcolberg, line 150: `std::optional` overloads [operator*](https://en.cppreference.com/w/cpp/utility/optional/operator*), which is preferable since it does not throw an exception.
```suggestion
      mmd_properties.push_back(*mem_id);
```
So the common optional pattern is to check whether a value is present, then safely access it using `operator*`.

src/acl_usm.cpp
bsyrowik: Oh really!? Interesting.  That is a bit odd.  Not sure why that would be.  I was just copying what we'd done in the device allocation code.  Should probably follow the spec if there is a good reason for it though!

pcolberg: Pardon, the correct specification to consider is the OpenCL extension, [cl_intel_mem_alloc_buffer_location which specifies a `cl_uint`](https://github.com/KhronosGroup/OpenCL-Docs/blob/46090155cc32baa68057d5f58a5bc7a302af3b33/extensions/cl_intel_mem_alloc_buffer_location.asciidoc?plain=1#L135-L137), so your current code is good. @tiwaria1, what are your thoughts on modifying the SYCL extension specification to match the OpenCL counterpart?

tiwaria1: > Lower bound of 0 is a good point. I changed it to cl_uint, so shouldn't need to worry about that any more. As for the upper limit, I don't think there is any way to enforce this. There may be a use-case for having arbitrary buffer locations? Not sure.

For Full System compiles the valid buffer_location range is `[0, NUM_GLOBAL_MEMORIES-1]`, where `NUM_GLOBAL_MEMORIES` can be inferred from the FPGA `board_spec.xml` file.

For IPA, the range is a unique set of positive integers >= `0`, eg: it can be `[0,1,2,3]` or `[1024,1025,1026]` etc. (Since these are ids that the user is assigning to their memory interfaces). This information is present in the `ipinterfaces.xml` file that we generate. For 2022.3 we can enforce a limit asking users to use positive integers starting from 0 and incrementing by 1 if that is absolutely needed.

bsyrowik:
Alright, I think we're good to go now.  @tiwaria1 @pcolberg - please let me know if there are any other changes you'd like :)

pcolberg:
Thanks @bsyrowik, looks good!

For `clDeviceMemAllocINTEL()`, commit f79a9c250979cb7818ecb95b5e5778ab4263e3c2 includes a unit test that queries and verifies the buffer location using `clGetMemAllocInfoINTEL()`. However, the buffer location is stored within a `cl_mem` struct, which is populated in `clDeviceMemAllocINTEL()` but not in `clHostMemAllocINTEL()` and `clSharedMemAllocINTEL()`. I wonder if we should add a `mem_id` field to `acl_usm_allocation_t` to persist the buffer location for later query?

pcolberg:
@bsyrowik I have made a slight modification to store the properties in an array on the stack similar to https://github.com/intel/llvm/pull/6220 and https://github.com/intel/llvm/pull/6218. Once this PR passes integration testing, this is ready to be merged.

PR #145, opened 2022-07-04
Pass streaming control signal names to simulator

pcolberg:
For the case of streaming control kernels in IP authoring, when passing
the start and done signals to the simulator MMD, instead of the kernel
interface name, pass the full signal names from the auto-discovery
string. Before this change, the runtime would pass the kernel name,
e.g., `_ZTSZ4mainE4MyIP` without the prefix including the kernel index
in case of multiple kernels. After this change, the runtime passes the
full signal names, e.g., `k0_ZTSZ4mainE4MyIP_streaming_start` including
the kernel index prefix and the streaming control signal suffix.

This amends https://github.com/intel/fpga-runtime-for-opencl/pull/103

Cc: @anandhv

### DIFF
--- a/include/acl_hal.h
+++ b/include/acl_hal.h
@@ -241,9 +241,9 @@ typedef struct {
                         mem_properties_t *properties, int *error);
 
   void (*simulation_streaming_kernel_start)(unsigned int physical_device_id,
-                                            const std::string &kernel_name);
+                                            const std::string &signal_name);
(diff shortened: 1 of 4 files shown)


zibaiwan (approved):
LGTM! Thanks Peter!

sophimao (approved):
Looks good! Thank you Peter!

Issue #147, opened 2022-07-14
Query whether the MMD supports buffer_location in shared/host USM allocations API calls


zibaiwan:
In this https://github.com/intel/fpga-runtime-for-opencl/pull/141, we pass the buffer location information from the OpenCL runtime to the MMD USM allocation APIs. However, there are some MMDs which don't support buffer_location property, and this caused the allocation fails on those MMDs. 

For 2022.3 release, the solution is to only pass buffer_location property to the MMD when targeting simulation. The query can be done through `acl_get_offline_device_user_setting`, 'CL_CONTEXT_MPSIM_DEVICE_INTELFPGA=1' or 'get_msim_mmd_layer()'.

For 2023.0, we could probably add something to the aocl_mmd_offline_info_t in aocl_mmd.h. Runtime could query whether the mmd supports buffer_location and only pass the buffer_location property if the mmd supports it. We need to test whether this approach breaks the existing MMDs that don't support buffer_locaiton, and also test it on the supporting MMD (probably starts with the simulation mmd). This requires co-ordination with the MMD owner.

ericxu233:
One small note: We currently don't cache the offline device user setting. So for this change, I have to query the environment variable again. We maybe want to add caching to these settings in the future?

sophimao:
Based on internal discussion the current plan is to use MMD version to determine whether buffer_location property is supported.

PR #149, opened 2022-07-15
Only pass buffer location property if in simulation environemnt

ericxu233:
This addresses [#147](https://github.com/intel/fpga-runtime-for-opencl/issues/147#issuecomment-1185725157).  
We only pass in the buffer location property if the design is in simulation environment. Also, we have to query the offline mode since that information is not cached anywhere.


src/acl_hal_mmd.cpp
```diff
@@ -2767,10 +2767,19 @@ void *acl_hal_mmd_host_alloc(const std::vector<cl_device_id> devices,
...
-      handles, devices.size(), size, alignment,
-      (aocl_mmd_mem_properties_t *)properties, error);
-  acl_delete_arr(handles);
+  int use_offline_only;
+  acl_get_offline_device_user_setting(&use_offline_only);
+
+  if (use_offline_only == ACL_CONTEXT_MPSIM) {
+    result = dispatch->aocl_mmd_host_alloc(
+        handles, devices.size(), size, alignment,
+        (aocl_mmd_mem_properties_t *)properties, error);
```
bsyrowik, line 2776: You may be able to reduce the amount of common code in the `if-else` by simply modifying `properties`, or using a sacrificial pointer that you set to `properties` if we're in sim mode?

ericxu233:
Yeah, this needs to be changed again. Just noticed that the if-else statement is actually wrong...

bsyrowik (commented):
Looks good, thanks Eric!

PR #153, opened 2022-08-10
kernel: add buffer allocations and migrations for streaming arguments

pcolberg:
This resolves a failure to pass local-memory sizes and global-memory
buffers as streaming arguments to a streaming kernel in simulation.

This amends https://github.com/intel/fpga-runtime-for-opencl/pull/103


zibaiwan (approved):
Thanks Peter!

PR #155, opened 2022-08-16
kernel: add buffer allocations and migrations for streaming arguments

pcolberg:
This backports https://github.com/intel/fpga-runtime-for-opencl/pull/153 to 2022.3.


zibaiwan (approved):
@pcolberg , looks good! Thank you for fixing this!

PR #157, opened 2022-09-12
Print Kernel hang status only in debug mode

zibaiwan:
In the earlier commit aeff9bbd95a22bf56b87912e6cc0b6192561e79c, this guarding conditions for the messages if (kern->io.debug_verbosity > 0) was removed accidentally, now adding it back.

Thanks @sophimao for discovering this regression.

Now the Runtime only prints the Kernel hang message in the debug mode.

### DIFF
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -1678,7 +1678,8 @@ void acl_kernel_if_check_kernel_status(acl_kernel_if *kern) {
 
   if (kern->last_kern_update != 0 &&
       (acl_kernel_if_get_time_us(kern) - kern->last_kern_update >
-       10 * 1000000)) {
+       10 * 1000000) &&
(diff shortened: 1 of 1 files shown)


zibaiwan:
> @zibaiwan Should this be backported to the 2022.3 branch? If so, after this is merged and you have the merged commit hash, please see the instructions for [backporting](https://github.com/intel/fpga-runtime-for-opencl/blob/1847ee9b08800b04e5d64b47ea362c780a5b21bd/RELEASE.md?plain=1#L69-L72) to submit a PR against 2022.3.

Thanks @pcolberg. I don't think this is required to be backported as there is no functional issue. The user will see their kernel status printed only if their kernel is hang. I believe it's fine to go into the next release.

sophimao (approved):
Looks good, thanks Zibai!

PR #162, opened 2022-09-14
Install ncurses-compat-libs in rockylinux-9-dev container

pcolberg:
This provides libtinfo.so.5 which is needed to run aocl-clang.

ncurses-compat-libs is provided in the devel repository, which provides
dependencies or devel packages that may not be provided by upstream.

https://wiki.rockylinux.org/rocky/repo/#notes-on-devel
https://bugzilla.redhat.com/show_bug.cgi?id=2117930

### DIFF
--- a/container/rockylinux-9-dev/Dockerfile
+++ b/container/rockylinux-9-dev/Dockerfile
@@ -18,6 +18,8 @@ COPY scripts/. ./
 RUN \
   sed -i '/^\[crb\]$/,/^\[/s#^enabled=0$#enabled=1#' /etc/yum.repos.d/rocky.repo \
   && grep '^enabled=1$' /etc/yum.repos.d/rocky.repo \
+  && sed -i '/^\[devel\]$/,/^\[/s#^enabled=0$#enabled=1#' /etc/yum.repos.d/rocky-devel.repo \
+  && grep '^enabled=1$' /etc/yum.repos.d/rocky-devel.repo \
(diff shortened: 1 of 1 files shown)


sophimao (approved):

container/rockylinux-9-dev/Dockerfile
```diff
@@ -18,6 +18,8 @@ COPY scripts/. ./
 RUN \
   sed -i '/^\[crb\]$/,/^\[/s#^enabled=0$#enabled=1#' /etc/yum.repos.d/rocky.repo \
   && grep '^enabled=1$' /etc/yum.repos.d/rocky.repo \
+  && sed -i '/^\[devel\]$/,/^\[/s#^enabled=0$#enabled=1#' /etc/yum.repos.d/rocky-devel.repo \
+  && grep '^enabled=1$' /etc/yum.repos.d/rocky-devel.repo \
```
sophimao, line 22: A quick question, what does the path `/etc/yum.repos.d/rocky-devel.repo` refer to? I guess I'm wondering where does that path point to?

sophimao: I see, does this path only exist in RHEL then? If I go on an RHEL machine and do `ls /etc/` then I will see `yum.repos.d`?

sophimao: Got it, thanks!

PR #172, opened 2022-10-20
Avoid changing ownership of workspace files in container workflows

pcolberg:
This avoids mismatching ownership of workspace files when running both container and non-container workflows on the same self-hosted runner.

Our self-hosted runners are configured to align the out-of-container runner user with the in-container non-root user. Changing the ownership of the workspace to the in-container root user makes them unwritable to the out-of-container runner user, which breaks non-container workflows.

### DIFF
--- a/.github/workflows/bdba.yml
+++ b/.github/workflows/bdba.yml
@@ -47,9 +47,6 @@ jobs:
     environment: bdba
 
     steps:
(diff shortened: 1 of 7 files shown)


zibaiwan (approved):
Thanks @pcolberg !

PR #173, opened 2022-10-20
Migrate workflows from deprecated set-output to environment file

pcolberg:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


zibaiwan (approved):
Thanks @pcolberg !

PR #174, opened 2022-10-20
Upgrade workflows to actions/cache@v3

pcolberg:
Resolves deprecation warning for `save-state` command.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


zibaiwan (approved):
Thanks @pcolberg !

PR #175, opened 2022-10-20
Upload binary package in build workflows

pcolberg:
Upload the default-installed components as a workflow artifact.


zibaiwan (approved):
Thanks Peter!

PR #177, opened 2022-10-20
Upgrade workflows to actions/checkout@v3

pcolberg:
Node.js 12 actions are deprecated.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

https://github.com/actions/checkout/blob/1f9a0c22da41e6ebfa534300ef656657ea2c6707/README.md#whats-new


zibaiwan (approved):
Thanks @pcolberg !

PR #182, opened 2022-10-24
Document Ubuntu 22.04 LTS support

pcolberg:


zibaiwan (approved):
Thanks @pcolberg !

PR #183, opened 2022-10-25
Move changelog entry for zlib to security section

pcolberg:


zibaiwan (approved):
Thanks @pcolberg

PR #186, opened 2022-10-25
Fix base image URL for klocwork-desktop-tools container

pcolberg:


zibaiwan (approved):
Thanks @pcolberg !

PR #189, opened 2022-10-27
Add background thread to update device op queue upon receiving kernel interrupt

sophimao:
Currently the runtime only makes update when:
- A synchronization API call (e.g., clFinish) is called
- An operation is being enqueued (e.g., clEnqueueNDRangeKernel)

This creates potential hang in the runtime. As there is limitation on FPGA that only one instance (launch) 
of the same kernel can be running on the board at one time, multiple launches has to be sequential, where 
later launches have to wait for the currently running launch to finish. For a kernel launch that has other 
launches waiting on it, when non of the two above update calls happens after that kernel launch finishes, 
the runtime will not submit waiting launches and thereby result in a hang.

In this change a background thread is added to update the device op queue when there is outstanding 
kernel interrupt to be served.

### DIFF
--- a/src/acl_platform.cpp
+++ b/src/acl_platform.cpp
@@ -78,6 +78,7 @@ static void l_initialize_devices(const acl_system_def_t *present_board_def,
                                  int offline_mode, unsigned int num_devices,
                                  const cl_device_id *devices);
 static void l_add_device(int idx);
+void *l_eagerly_update_device_op_queue(void *arg);
 
 //////////////////////////////
 // OpenCL API
@@ -412,6 +413,10 @@ void acl_init_platform(void) {
 
   // Device operation queue.
   acl_init_device_op_queue(&acl_platform.device_op_queue);
+  // Send off device_op_queue update thread
+  acl_platform.outstanding_interrupt = 0;
+  acl_thread_create(&acl_platform.device_op_queue_update_thread, 0,
+                    l_eagerly_update_device_op_queue, NULL);
 
   // Initialize sampler allocator.
   for (int i = 0; i < ACL_MAX_SAMPLER; i++) {
@@ -735,6 +740,25 @@ static void l_add_device(int idx) {
 
   device->min_local_mem_size = 16 * 1024; // Min value for OpenCL full profile.
   device->address_bits = 64;              // Yes, our devices are 64-bit.
+}
+
+void *l_eagerly_update_device_op_queue(void *arg) {
+  while (true) {
+    std::scoped_lock lock{acl_mutex_wrapper};
+
+    // Sleep if no interrupt happening
+    acl_wait_for_device_update(NULL);
+
+    if (!acl_platform.initialized) {
+      break;
+    }
+    if (acl_platform.outstanding_interrupt) {
+      acl_print_debug_msg("Serving outstanding kernel interrupt...\n");
+      acl_update_device_op_queue(&(acl_platform.device_op_queue));
+      acl_platform.outstanding_interrupt = 0;
+    }
+  }
+  return NULL;
 }
 
 // These functions check to see if a given object is known to the system.
--- a/src/acl_thread.cpp
+++ b/src/acl_thread.cpp
@@ -9,6 +9,7 @@
 #include <acl_context.h>
 #include <acl_hal.h>
 #include <acl_thread.h>
+#include <acl_util.h>
 
 ACL_TLS int acl_global_lock_count = 0;
 ACL_TLS int acl_inside_sig_flag = 0;
@@ -55,7 +56,7 @@ void acl_mutex_wrapper_t::resume_lock(int lock_count) {
 
 void acl_wait_for_device_update(cl_context context) {
   acl_assert_locked();
-  if (acl_get_hal()->get_debug_verbosity &&
+  if (acl_context_is_valid(context) && acl_get_hal()->get_debug_verbosity &&
       acl_get_hal()->get_debug_verbosity() > 0) {
     unsigned timeout = 5; // Seconds
     // Keep waiting until signal is received
@@ -102,6 +103,12 @@ __attribute__((constructor)) static void l_global_lock_init() {
 }
 
 __attribute__((destructor)) static void l_global_lock_uninit() {
+  {
+    std::scoped_lock lock{acl_mutex_wrapper};
+    acl_get_platform()->initialized = 0;
+    acl_signal_condvar(&l_acl_global_condvar); // wake up waiting thread
+  }
+  acl_thread_join(&acl_get_platform()->device_op_queue_update_thread);
   acl_reset_condvar(&l_acl_global_condvar);
 }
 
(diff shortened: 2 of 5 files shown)


src/acl_thread.cpp
```diff
@@ -102,6 +103,12 @@ __attribute__((constructor)) static void l_global_lock_init() {
 }
 
 __attribute__((destructor)) static void l_global_lock_uninit() {
+  {
+    std::scoped_lock lock{acl_mutex_wrapper};
+    acl_get_platform()->initialized = 0;
+    acl_signal_condvar(&l_acl_global_condvar); // wake up waiting thread
+  }
+  acl_thread_join(&acl_get_platform()->device_op_queue_update_thread);
```
sophimao, line 111: Not sure if this section of code is necessary here? The situation I have in mind is when there is an interrupt coming and waking up the device op queue update thread, and the program is about to unload the library, then the thread will acquire the lock and get preempted, when `acl_reset_condvar` is called it will see a locked lock, and that might result in some problem. But I guess having an interrupt coming in when the program is about to end is pretty rare?

src/acl_platform.cpp
```diff
@@ -737,6 +742,25 @@ static void l_add_device(int idx) {
   device->address_bits = 64;              // Yes, our devices are 64-bit.
 }
 
+void *l_eagerly_update_device_op_queue(void *arg) {
+  while (true) {
+    std::scoped_lock lock{acl_mutex_wrapper};
+
+    // Sleep if no interrupt happening
+    acl_wait_for_device_update(NULL);
```
sophimao, line 750: The side effect of using this existing call is that the thread gets woken up not just when there is a mmd triggered kernel interrupt, but also when all the signal handler functions are called (`acl_receive_kernel_update`, `acl_set_execution_status`, `acl_receive_device_exception`, and `acl_schedule_printf_buffer_pickup`). But this is already better then always polling, I wonder if it is necessary to fine grain this even further (might be risky)?

sophimao:
Got a thread leak... Will look into this further.

src/acl_thread.cpp
sophimao: Based on the unit tests it seems like this section might still be required, even with this joining code some of the CI jobs are failing when exiting the program:
```
6: TEST(auto_configure, many_ok_forward_compatibility) - 22 ms
6: TEST(auto_configure, simple) - 0 ms
6: 
6: OK (224 tests, 224 ran, 138598587 checks, 0 ignored, 0 filtered out, 14787 ms)
6: 
6: acl_test: ../lib/acl_threadsupport/src/acl_threadsupport.c:372: acl_release_condvar: Assertion `ret == 0' failed.
6/6 Test #6: acl_test .........................Child aborted***Exception:  21.01 sec
```

sophimao:
Performance testing using regtest/hld/sycl/microbenchmarks/runtime/kernel_latency:
- a10_ref: Single kernel execution time: `30.8214 us` -> `37.7533 us`
- pac_a10: Queueing

This is not good 😞

sophimao:
Performance testing using kernel_latency test:
- a10_ref: Single kernel execution time: `30.8214 us` -> `37.7533 us` (This is not good 😞)
- pac_a10: Some issue with kernel interface version id and I can't run the design 😦
Error: `HAL Kern: Version mismatch! Expected 0xa0c00001 but read 0x0`

sophimao:
On synchronization between signal handler and other user threads, seems like the only safe primitives to use inside linux signal handler are:
- `sig_atomic_t`
- lock-free [`std::atomic`](https://en.cppreference.com/w/cpp/atomic/atomic)

In c++20, `std::atomic` include conditional wait calls `wait`, `notify_one` and `notify_all`

Potential ways to synchronize with signal handler are:
- Use semaphores (comes with c++20 in the c++ concurrency library)
- Self-pipe trick (linux only)
- Use `sigwait`/`sigwaitinfo` (linux only)
- Use `signalfd` (linux only)
- lock-free `std::atomic` (must poll without the wait and notify calls)
- `sig_atomic_t` (must poll)

src/acl_platform.cpp
sophimao: Can potentially
- Introduce a new `acl_condvar_s`, or
- Use a semaphore (is this doable?)
To signal to the device op queue update thread, to avoid waking up from event status update calls, etc. 

However, the device op queue update thread might still need to acquire the global lock when doing the update, so that no other thread is modifying the shared resources (which are quite a lot...).

src/acl_platform.cpp
```diff
@@ -737,6 +742,25 @@ static void l_add_device(int idx) {
   device->address_bits = 64;              // Yes, our devices are 64-bit.
 }
 
+void *l_eagerly_update_device_op_queue(void *arg) {
+  while (true) {
+    std::scoped_lock lock{acl_mutex_wrapper};
```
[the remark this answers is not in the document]
sophimao: When the thread waits it releases the lock so that other threads can make progress, are you referring to the time between the lock and wait during which other threads will be blocked?

src/acl_platform.cpp
sophimao: Thanks for the feedback! Yes, the primary goal is to make forward progress in processing the device operation queue. In the model you suggested, is it kind of like treating the device op queue as a buffer for consumer (runtime spawned thread) and producer (user threads)? This conceptually makes sense, just that I'm not sure if there are other runtime constructs (events, command queue, etc.) modified during the process of updating the device op queue, would that pose any obstacle to the consumer-producer model?

Also I'm thinking, if at the end of the day we are actively making progress on the device op queue, maybe it would be possible to remove some of the `acl_idle_updates` calls in `clEnqueue...` used to nudge the device op queue. I'm not sure if this idea is practical in any sense or if it will cause any issue though...

src/acl_globals.cpp
```diff
@@ -231,6 +231,35 @@ void acl_reset(void) {
...
+    std::scoped_lock lock{acl_mutex_wrapper};
+
+    l_reset_present_board();
+
+    acl_platform.offline_device = "";
+    acl_platform.num_devices = 0;
+    for (unsigned i = 0; i < ACL_MAX_DEVICE; ++i) {
+      acl_platform.device[i] = _cl_device_id();
+    }
+    acl_platform.initialized = 0;
```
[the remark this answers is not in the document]
sophimao: I remember seeing some issue if I call `acl_signal_device_update` without the lock or in another critical section, but I can't remember clearly if that's the case. I'll try to think through it again to see if using `acl_reset` here is okay.

src/acl_thread.cpp
```diff
@@ -102,6 +103,14 @@ __attribute__((constructor)) static void l_global_lock_init() {
...
 
 __attribute__((destructor)) static void l_global_lock_uninit() {
+  if (acl_get_platform()->device_op_queue_update_thread) {
+    {
+      std::scoped_lock lock{acl_mutex_wrapper};
+      acl_get_platform()->initialized = 0;
+      acl_signal_condvar(&l_acl_global_condvar); // wake up waiting thread
+    }
+    acl_thread_join(&acl_get_platform()->device_op_queue_update_thread);
+  }
```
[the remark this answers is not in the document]
sophimao: I originally included the thread as a member of `acl_platform` so I created it when initializing the platform. However the `acl_reset` function is never called in a user flow so I had to put it in the destructor... I agree that it's probably better to do this when creating and releasing context, but need to add some flag indicating the first creation and the last release.

PR #196, opened 2022-11-07
acl_profiler_test: fix undefined behaviour

pcolberg:
This resolves the final issue before [enabling address sanitizer](https://github.com/intel/fpga-runtime-for-opencl/pull/118) :partying_face:

`putenv()` should be replaced with `acl_test_setenv()` and `acl_test_unsetenv()` in a subsequent pull request, but for now this is sufficient to pass address sanitizer.


zibaiwan (approved):
Thanks @pcolberg !

PR #199, opened 2022-11-15
acl_hash: fix integer conversion warning

pcolberg:
Avoid integer conversion altogether by using a more efficient
implementation which compiles to branchless code with a rotate
instruction, assuming the number of bits to rotate by is known
at compile time so the compiler may eliminate the assert().


zibaiwan (approved):
Thanks @pcolberg for fixing this!

PR #200, opened 2022-11-16
Fix CSR query issue to enable new CSR protocal

ericxu233:
In hardware runs, it seems that there are some dummy kernels present that messes up with the CSR check (the dummy kernels have a csr version of 4). When the runtime checks the csr for the dummy kernels, they will overwrite the default cra_address_offset causing errors with the new compiler change. Adding this "else" statement will make sure that the cra_address_offset is set correctly.


ericxu233:
Newest sycl-l3 run shows no additional errors: https://spetc.intel.com/testsummary?testRunIds=7216415

ericxu233:
@pcolberg @zibaiwan I've described the underlying root cause of the issue in the main comment above. This else statement would suffice in fixing the issue. There might be some runtime deficiencies where we create contexts and query the auto-discovery  string and CSR version too many times but that would require more investigation and likely out of scope for this change. This change is ready for review now.

zibaiwan (changes requested):
Thanks @ericxu233 . I think the `else` statement makes sense. I would suggest add a bit more detail in the code comments to explain why the old CSR version can be possibly queried, because the Runtime would call try_device to load the default aocx or whatever aocx was programmed on the board before.

zibaiwan (approved):
Thanks @ericxu233 !

PR #205, opened 2022-11-21
workflows: add Coverity static code analysis

pcolberg:
For now, the workflow only lists the issues. Once the Coverity issues are
resolved, the output of cov-format-errors can be tested for zero length.


sophimao (approved):
Looks good. Thanks Peter!

PR #208, opened 2022-11-22
workflows: upgrade Klocwork installation

pcolberg:
Switch from Debug to Release build, which has no impact on the analysis.
    

zibaiwan (approved):
Thanks @pcolberg !

PR #211, opened 2022-11-23
workflows: use ccache to speed up code coverage

pcolberg:
ccache requires the build paths, e.g., for include directories to
be identical in order to reuse the cache between the two builds.

Setting the CCACHE_BASEDIR environment variable to the respective build
directory does not seem to have any effect and does not reuse the cache.


zibaiwan (approved):
Thanks @pcolberg !

PR #212, opened 2022-11-23
container: install ccache

pcolberg:
ccache is not available on Rocky Linux.


zibaiwan (approved):
Thanks @pcolberg !

PR #213, opened 2022-11-24
Backport workflow fixes to 2023.0

pcolberg:
This fixes the clang-format and klocwork workflows whose custom containers have been removed. The clang-format workflow could run when the release is tagged. The klocwork pull-request workflow would run if a source code change were backported. The ccache change is not strictly needed but included anyway to minimize the divergence from the main branch infrastructure.


zibaiwan (approved):
Thanks @pcolberg !

PR #215, opened 2022-11-25
pkg_editor: unconditionally null-terminate output of strncpy()

pcolberg:
Resolves Coverity error "Buffer not null terminated (BUFFER_SIZE)".

The previous code was correct but non-idiomatic.

@haoxian2 in case you see this pattern elsewhere.


zibaiwan (approved):
Thanks @pcolberg !

PR #218, opened 2022-11-30
Fixing coverity issues in src/acl_usm.cpp

haoxian2:
Fixing `Type: AUTO_CAUSES_COPY (AUTO_CAUSES_COPY)`
Error caused because `auto` in `for-each` loop copies the iterator object from the array/list.
Adding `&` to reference the original element.
Adding `const` so the code don't mistakenly modify the object.


haoxian2:
@pcolberg I changed the commit message to the PR description. Thank you for the suggestion!

PR #226, opened 2022-12-08
Fixed coverity issue in acl_kernel_if.cpp

haoxian2:
Continuation of discussion featured in https://github.com/intel/fpga-runtime-for-opencl/pull/220.
Fixed coverity issue in acl_kernel_if.cpp: Type: Reliance on integer endianness (INCOMPATIBLE_CAST)

In the three instances where this issue happens, the function acl_kernel_if_read_32b is used at the end of the scope to query about what's inside the acl_kernel_if* kern object. However, they don't do anything with the obtained information. I believe this read operation is performed to confirm that the acl_kernel_if* kern has been written properly. So if the read operation successed, then the write operation must have also succeeded as well. The functions employing acl_kernel_if_read_32b are used to return the segment offset. This segment offset is then used to perform another read/write, which return status would then be used for error checking. Therefore, the `acl_kernel_if_read_32b` function is redundant and can be removed from the aformentioned 3 instances.

### DIFF
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -453,8 +453,6 @@ static uintptr_t acl_kernel_cra_set_segment(acl_kernel_if *kern,
     acl_kernel_if_write_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
                             (unsigned int)segment);
     kern->cur_segment = segment;
-    acl_kernel_if_read_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
-                           (unsigned int *)&segment);
   }
 
   return segment_offset;
@@ -470,8 +468,6 @@ static uintptr_t acl_kernel_cra_set_segment_rom(acl_kernel_if *kern,
     acl_kernel_if_write_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
                             (unsigned int)segment);
     kern->cur_segment = segment;
-    acl_kernel_if_read_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
-                           (unsigned int *)&segment);
   }
 
   return segment_offset;
@@ -1613,8 +1609,6 @@ void acl_kernel_if_update_status(acl_kernel_if *kern) {
     acl_kernel_if_write_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
                             (unsigned int)segment_pre_irq);
     kern->cur_segment = segment_pre_irq;
-    acl_kernel_if_read_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
-                           (unsigned int *)&segment_pre_irq);
   }
 }
 


haoxian2:
The function acl_kernel_if_read_32b does r = kern->io.read(&kern->io, (dev_addr_t)addr, (char *)val, (size_t)size) as a read of val and if r < size then it does a print and return -1 kern->io.printf( "HAL Kern Error: Read failed from addr %x, read %zu expected %zu\n", addr, r, size). So technically it doesn't really do anything except printing out information if the read somehow failed. Therefore, the val doesn't need to be initialized.

haoxian2:
After some investigation, I found the following:
- The function in question in 'r = kern->io.read(&kern->io, (dev_addr_t)addr, (char *)val, (size_t)size)' is 'kern->io.read' which is found in `acl_bsp_io.h`. It is found in the form of `size_t (*read)(acl_bsp_io *io, dev_addr_t src, char *dest, size_t size)`
- It's a function pointer, but I can't seem to find the definition of where it is pointing to.
@pcolberg @zibaiwan Do you know where this pointer is pointing to?

zibaiwan:
> After some investigation, I found the following:
> 
> * The function in question in `r = kern->io.read(&kern->io, (dev_addr_t)addr, (char *)val, (size_t)size)` is `kern->io.read` which is found in `acl_bsp_io.h`. It is found in the form of `size_t (*read)(acl_bsp_io *io, dev_addr_t src, char *dest, size_t size)`
> * It's a function pointer, but I can't seem to find the definition of where it is pointing to.
>   @pcolberg @zibaiwan Do you know where this pointer is pointing to?

Hi @haoxian2 , I think it's [ `acl_kernel_if_read`](https://github.com/intel/fpga-runtime-for-opencl/blob/23423327dde22b9613d09e64b559ac0c25b7625e/src/acl_hal_mmd.cpp#L1026).

haoxian2:
After a bit of digging, I concluded that `acl_kernel_if_read_32b` only does a read by only modifying the `segment` variable. It is employed to confirmed that the previous write operation executed successfully. However, this is redundant since the function that employs `acl_kernel_if_read_32b` are used to return a `segment_offset` which is then used in a read/write operation again, but this time, there is an error checking in place for if the operation was successful. Since the subsequent read/write employs the `segment_offset` return value, it will read the same information as the previous `acl_kernel_if_read_32b`, making it redundant. Thus, I removed it.

zibaiwan (commented):
Thanks Hao for fixing this. I need a bit more time to review this.

zibaiwan (commented):
@haoxian2 , I reviewed the very first swarm change that introduced the following two lines. It mainly intended to add the `kern->cur_segment = segment;`, the rest of two occurrences were just copied from this code.

```
 kern->cur_segment = segment;
 acl_kernel_if_read_32b(kern, OFFSET_KERNEL_CRA_SEGMENT,
                           (unsigned int *)&segment);
```

I agree with you analysis that the `acl_kernel_if_read_32b` can be deleted. The read is going into the local `segment` variable which is not propagated to anywhere. 

In case we miss anything, can you please rebase your branch to the latest runtime codebase and open a PR against acl repo to do the integration testing?

Thanks,

Zibai

haoxian2:
@zibaiwan I've opened a PR on acl repo containing this runtime change: https://github.com/intel-innersource/applications.fpga.oneapi.products.acl-source/pull/680 

It seems like everything is passing

zibaiwan:
@haoxian2 , thanks!

zibaiwan (approved):

PR #228, opened 2022-12-09
Fixed coverity issues in acl_program.cpp

haoxian2:
# Fixed coverity in acl_program.cpp: Dead default in switch (DEADCODE)
(line 256) Default branch never reached since `switch (pass)` only has 0 and 1 as values because `pass` is defined in a for-loop (line 225)  as `< 2`.

# Fixed coverity in acl_program.cpp: Dereference before null check (REVERSE_INULL)
(line 145) Program is already dereferenced in the for-loop above to check `program->num_devices`, so the check would also pass. 

# Fixed coverity in acl_program.cpp: Use after free (USE_AFTER_FREE)
(line 1834) Coverity complains that you should not print the pointer address value if it has already been freed. Since we are not accessing what's in the pointer itself, it is safe to query about pointer address value. Therefore, instead of passing the pointer as the argument, a dereferenced pointer to that pointer is passed as an argument instead.

### DIFF
--- a/src/acl_program.cpp
+++ b/src/acl_program.cpp
@@ -1819,6 +1819,9 @@ static cl_int l_build_from_source_in_dir(acl_device_program_info_t *dev_prog,
   if (status == CL_SUCCESS && context->compiles_programs) {
     for (unsigned i = 0; i < num_build_cmds; i++) {
       struct acl_file_handle_t *cmdfp = acl_fopen(cmdfile[i], "wb");
+      if (debug_mode > 0) {
+        printf(" cmdfp %p\n", cmdfp);
+      }
       size_t bytes_written = 0;
       if (cmdfp) {
         bytes_written = (size_t)acl_fprintf(cmdfp, "%s", cmd[i].c_str());
@@ -1826,7 +1829,6 @@ static cl_int l_build_from_source_in_dir(acl_device_program_info_t *dev_prog,
       }
       if (!cmdfp || (cmd[i].length() != bytes_written)) {
         if (debug_mode > 0) {
-          printf(" cmdfp %p\n", cmdfp);
           printf(" cmdlen %d  bw %d\n", (int)cmd[i].length(),
                  (int)bytes_written);
         }
--- a/src/acl_program.cpp
+++ b/src/acl_program.cpp
@@ -142,8 +142,7 @@ CL_API_ENTRY cl_int CL_API_CALL clReleaseProgramIntelFPGA(cl_program program) {
       if (program->device[i]->last_bin != nullptr)
         program->device[i]->last_bin->unload_content();
     }
-    if (program)
-      l_free_program(program);
+    l_free_program(program);
   }
   return CL_SUCCESS;
 }


src/acl_program.cpp
```diff
@@ -142,8 +142,7 @@ CL_API_ENTRY cl_int CL_API_CALL clReleaseProgramIntelFPGA(cl_program program) {
       if (program->device[i]->last_bin != nullptr)
         program->device[i]->last_bin->unload_content();
     }
-    if (program)
-      l_free_program(program);
+    l_free_program(program);
```
zibaiwan, line 145: Thanks @haoxian2. I checked and indeed this if(program) should be removed here.

src/acl_program.cpp
```diff
@@ -1824,14 +1821,15 @@ static cl_int l_build_from_source_in_dir(acl_device_program_info_t *dev_prog,
...
+      auto cmdfp_addrs = &cmdfp;
       size_t bytes_written = 0;
       if (cmdfp) {
         bytes_written = (size_t)acl_fprintf(cmdfp, "%s", cmd[i].c_str());
         acl_fclose(cmdfp); // close in any case.
       }
       if (!cmdfp || (cmd[i].length() != bytes_written)) {
         if (debug_mode > 0) {
-          printf(" cmdfp %p\n", cmdfp);
+          printf(" cmdfp %p\n", (void *)*cmdfp_addrs);
```
zibaiwan, line 1832: Looking into the coverity log.

`acl_fclose`, which calls `fclose` frees `cmdfp`. Coverity is complaining Passing freed pointer `cmdfp` as an argument to "printf".

I am not 100% sure about Hao's solution. I think it's kind of just make Coverity happy and will not print useful data when the `cmdfp` is indeed freed?

I think maybe we should use something like

`if (debug_mode > 0 && cmdfp) printf(" cmdfp %p\n", cmdfp);`

@haoxian2 , what do you think? Let me know if you want Ilan to take it over.

haoxian2: @zibaiwan Thank you for looking into it! 
I'm questioning the relevance of this if statement: `if (!cmdfp || (cmd[i].length() != bytes_written))`

If `cmdfp` is valid, then the previous if statement block `if (cmdfp)` would call `acl_fclose(cmdfp)` to make it invalid. Therefore, when arriving at `if (!cmdfp || (cmd[i].length() != bytes_written))`, `cmdfp` would always be false and `!cmdfp` would logically be always true.
In this case `if (debug_mode > 0 && cmdfp) printf(" cmdfp %p\n", cmdfp);` statement would never execute.

I have two ideas why this debug printf is implement.
1. Since `cmdfp` is a pointer. Printing it would check the address of whatever it was pointing to. Freeing the pointer, in my understand, is just telling us that that memory is unlocked to other processes and does not replace the address saved in the pointer unless it was set to NULL afterwards. Thus, that printf statement prints the address saved in the pointer for whatever reason. If that's the case, we can remove this printf statement if this information is not important for debugging anymore.
2. This printf statement is misplaced and should be in the preceding `if (cmdfp)` if statement.

In any case, I think we should remove this `if (!cmdfp || (cmd[i].length() != bytes_written))` since it will always be true.

What are your thoughts on this subject matter?

zibaiwan: @haoxian2, thanks for looking into this further.

Regarding your logic "I think we should remove this if (!cmdfp || (cmd[i].length() != bytes_written)) since it will always be true.", your point definitely makes sense. However, I suspect, `acl_fclose(cmdfp)` doesn't actually make `!cmdfp` to be evaluated to true.

Because if `!cmdfp` is always true here, I think this function will always fail with `CL_BUILD_PROGRAM_FAILURE` error. I think we test building program from source somewhere, so this shouldn't be the case, **hopefully**.

```
if (!cmdfp || (cmd[i].length() != bytes_written)) {
        if (debug_mode > 0) {
          printf(" cmdfp %p\n", cmdfp);
          printf(" cmdlen %d  bw %d\n", (int)cmd[i].length(),
                 (int)bytes_written);
        }
        status = CL_BUILD_PROGRAM_FAILURE;
        acl_context_callback(context, "Can't write command file: ");
        acl_context_callback(context, acl_support_last_error());
      }
```
You can maybe try to do some testing, such as writing a simple C++ program with file handler and see if fclose that file handler would make (!Filehandler) to be evaluated to True?

To solve this particular issue, I am thinking since those are just debug messages, can we just move that print before the `fclose` call is made? I think we just print the address after we open the file for debug purpose.

zibaiwan (changes requested):

src/acl_program.cpp
haoxian2: @zibaiwan I tested in a simple C++ program with file handler and saw that even after using fclose(), !Filehandler evaluated to False and skipped the if-statement block. Therefore, I agree with your solution to move that print before fclose call is made.

zibaiwan (commented):
Thanks @haoxian2! Looks good to me, can you please fix the clang format, then we can merge this!

zibaiwan (dismissed):

zibaiwan (approved):
Thanks @haoxian2 !

PR #230, opened 2022-12-13
Enabled ACL_SUPPORT_DOUBLE to support cl_khr_fp64

haoxian2:


zibaiwan:
Due to this SYCL change https://github.com/intel/llvm/pull/6982, this is a critical issue. Because if we don't solve this, we will lose double support in oneAPI. This change should go into 2023.1. Hao already tested the previous failed tests in simulation and hardware.

zibaiwan (approved):
Thanks @haoxian2 !

PR #232, opened 2022-12-19
workflows: only run single instance of docker prune at a time

pcolberg:
This resolves a workflow failure when two jobs are queued while the runnner is offline and start concurrently when it goes online again.


sophimao (approved):
Thanks Peter!

PR #234, opened 2023-01-04
Cache context offline device setting specified by environment variables

sophimao:
The context offline mode is set through environment variables and remains constant throughout the host program lifetime. This change caches the context offline device setting in the platform which saves the overhead of calling the environment variable querying function.


sophimao:
Thanks Peter for the review! I actually have a separate change just targeting the QoR regression introduced in #214, just that I didn't include the enum change in there. I will modify that and create another PR, and change this PR to a draft.

PR #236, opened 2023-01-05
Add copy constructor to acl_device_binary_t

IlanTruanovsky:
Add default copy constructor to acl_device_binary_t to fix a Coverity issue.


zibaiwan:
> @IlanTruanovsky It's been a while, but I submitted a fix for the same issue in #207 and ran into [this Klocwork issue](https://github.com/intel/fpga-runtime-for-opencl/pull/207#issuecomment-1374130754). Klocwork's analysis is flawed here since the returned value is a reference into `m_split_device_binaries`, not a reference to a local variable. Since Klocwork is being phased out, we should go ahead and merge despite the Klocwork false positive. @zibaiwan, what do you think?

Thanks @pcolberg for catching this false positive. I agree we should merge it.

PR #237, opened 2023-01-06
Remove dead code at line 256 of acl_program.cpp

intel-liudean:


zibaiwan (approved):
Thanks @intel-liudean !

PR #238, opened 2023-01-09
Change size_t printf statement to %zu

IlanTruanovsky:
This change fixes this Coverity issue:
src/acl_offline_hal.cpp:182:69:
  Type: Invalid type in argument to printf format specifier (PRINTF_ARGS)


zibaiwan (changes requested):
Thanks Ilan, based on the resolved discussion in https://github.com/intel/fpga-runtime-for-opencl/pull/221.

We should use `%zu` for the `size_t` type.

zibaiwan (approved):

PR #244, opened 2023-01-13
Remove CL_CONTEXT_COMPILER_MODE_SIMULATION_INTELFPGA

intel-liudean:
This compiler mode (mode 6) is unused by the runtime, safe to remove.

Tests Done:
Runtime Unit Tests
Checked that products/regtest/hld does not have lines like "compiler_mode => 6" or "compiler_mode => '6'"


sophimao (dismissed):
Thanks Dean! Looks good to me :)

Another thing is that I tried search for the compiler mode on Google but nothing returned, so it's also not documented anywhere. Not sure if this is worth mentioning in the PR description?

intel-liudean:
> compiler mode on Google

I will mention it, I think there is no down side by mentioning it.

sophimao:
Thanks @pcolberg for the comment! It appears to me that the compiler mode here is only used in runtime though, unless I'm missing something? But you are right, it's probably hard to look for those with public searches.

intel-liudean:
I have included a description in the commit message for 1) why I am making this change and 2) what tests I have done to make sure no regression happens.

intel-liudean:
Thank you @pcolberg for merging this change!

PR #248, opened 2023-01-16
Add `dlclose` to prevent memory leak

IlanTruanovsky:
A dynamically allocated pointer was not being properly cleaned up.

I think the ideal way of handling this (if this was in C++) would be to create a "library" class with proper constructors/deconstructions. Is there a particular reason why this is written in C?

This fixes the following Coverity issue:
```
lib/pkg_editor/src/zlib.c:119:1:
  Type: Resource leak (RESOURCE_LEAK)

lib/pkg_editor/src/zlib.c:98:3:
  1. path: Condition "zlib_interface.deflateInit_", taking false branch.
lib/pkg_editor/src/zlib.c:101:3:
  2. alloc_fn: Storage is returned from allocation function "zlib_load_library".
lib/pkg_editor/src/zlib.c:90:3:
  2.1. path: Condition "dlopen_handle", taking false branch.
lib/pkg_editor/src/zlib.c:93:3:
  2.2. alloc_fn: Storage is returned from allocation function "dlopen".
lib/pkg_editor/src/zlib.c:93:3:
  2.3. return_alloc_fn: Directly returning storage allocated by "dlopen".
lib/pkg_editor/src/zlib.c:101:3:
  3. var_assign: Assigning: "library" = storage returned from "zlib_load_library()".
lib/pkg_editor/src/zlib.c:102:3:
  4. path: Condition "library == NULL", taking false branch.
lib/pkg_editor/src/zlib.c:110:3:
  5. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  5.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:111:3:
  6. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  6.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:112:3:
  7. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  7.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:113:3:
  8. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  8.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:114:3:
  9. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  9.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:115:3:
  10. noescape: Resource "library" is not freed or pointed-to in "load_one_symbol".
lib/pkg_editor/src/zlib.c:39:36:
  10.1. noescape: "load_one_symbol(void *, char const *)" does not free or save its parameter "library".
lib/pkg_editor/src/zlib.c:119:1:
  11. leaked_storage: Variable "library" going out of scope leaks the storage it points to.
```


lib/pkg_editor/src/zlib.c
```diff
@@ -116,6 +116,11 @@ static void zlib_load() {
 #ifdef _MSC_VER
 #pragma warning(pop)
 #endif
+#ifdef _WIN32
+  FreeLibrary(library)
```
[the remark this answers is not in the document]
IlanTruanovsky: Yes, this is all that's needed to fix the compilation error.

I've implemented this in my local version of the repo and have been trying to figure out why it's been failing unit tests before updating the PR.

lib/pkg_editor/src/zlib.c
```diff
@@ -116,6 +116,11 @@ static void zlib_load() {
 #ifdef _MSC_VER
 #pragma warning(pop)
 #endif
+#ifdef _WIN32
+  FreeLibrary(library);
```
zibaiwan, line 120: I am not sure if this caused the segfault, but I think it should be `FreeLibrary((HMODULE)library);`, `zlib_load_library` returned a `void *` but `FreeLibrary` expected a `HMODULE`

IlanTruanovsky: I believe what is happening is that calling `dlclose` assumes that we are done with the library and so frees the memory, but does not replace that area in memory with invalid data. So, calling the functions from the DLL still work, although it's not being used as intended. (https://pubs.opengroup.org/onlinepubs/7908799/xsh/dlclose.html)

FreeLibrary probably does invalidate the data where the pointer pointed to when called, although I could not find any documentation explicitly mentioning this.

IlanTruanovsky:
> @IlanTruanovsky Could you summarize your understanding of the Coverity error in words so we can figure out what needs to be done here?

Coverity is complaining about the `library` variable going out of scope in the `zlib_load` function. The library variable was loaded in memory using `dlopen`, and since it goes out of scope, it should be unloaded to prevent a memory leak.

@pcolberg, you made an interesting point in one of the above conversations. We shouldn't be unloading the library, even if it goes out of scope. There should be some other solution.

> Is the issue only in a non-default, error flow? Or is the zlib handle not unloaded at all, in the default flow?

The issue would happen in any flow that uses zlib (not sure if this the default or non-default flow). The library is never explicitly 
 unloaded using any function call.

There must be a simple way to free the library after having called the symbols loaded in from the library. I'll look over it some more and see if I can come up with something.

PR #249, opened 2023-01-19
auto_configure: parse device global address in base 16

sophimao:
The compiler use a base 16 number as the device global address in the autodiscovery string, this change matches the runtime autodiscovery parsing with the compiler generated string.


sophimao:
> Would reading the address as base 16 not break compatibility?

I don't think it will, as the compiler always generates a base 16 number, even in the 2023.1 release. The reason why there is no test failures or anything is because this value currently is not used any where during host program run.

PR #251, opened 2023-01-24
pkg_editor: fix build when zlib is not available

pcolberg:
This amends https://github.com/intel/fpga-runtime-for-opencl/pull/248


zibaiwan (approved):
Thanks @pcolberg !

PR #254, opened 2023-01-26
Mark COPY_PASTE_ERROR as false positives

IlanTruanovsky:
After taking a look at the code, it does not appear that these are copy paste errors.

I don't really like putting a comment to mark Coverity issues as false-positives, but I dislike obfuscating the code to resolve the error even more.

Fixes:
```
test/acl_device_op_test.cpp:987:3:
  Type: Copy-paste error (COPY_PASTE_ERROR)

test/acl_device_op_test.cpp:984:3:
  original: "op0->timestamp" looks like the original copy.
test/acl_device_op_test.cpp:987:3:
  copy_paste_error: "op0" in "op0->timestamp" looks like a copy-paste error.
test/acl_device_op_test.cpp:987:3:
  remediation: Should it say "op1" instead?

test/acl_device_op_test.cpp:906:3:
  Type: Copy-paste error (COPY_PASTE_ERROR)

test/acl_device_op_test.cpp:878:3:
  original: "op0->timestamp" looks like the original copy.
test/acl_device_op_test.cpp:906:3:
  copy_paste_error: "op0" in "op0->timestamp" looks like a copy-paste error.
test/acl_device_op_test.cpp:906:3:
  remediation: Should it say "op1" instead?
```


IlanTruanovsky:
> > I don't really like putting a comment to mark Coverity issues as false-positives, but I dislike obfuscating the code to resolve the error even more.
> 
> Yes, I agree with both. Maybe [swapping the arguments works as suggested here](https://github.com/intel/fpga-runtime-for-opencl/pull/256#discussion_r1088403732)? Besides swapping arguments, does reordering these lines have any effect? Taking a step back, how many more of these copy-paste issues are you seeing? Is there a common pattern to these issues that triggers Coverity?

There are only 5 of these Coverity issues in the runtime. I believe what triggers this is having similar expressions for some amount of lines, and then one line suddenly changing one (or possibly more) of the variables involved in those expressions, but utilizing the same overall structure. Reordering the lines does work, but I don't like that we have to do this just to avoid Coverity from complaining. It is still better than adding a comment, though.

> Are all of these false positives?

It looks like 4/5 of them are. The only one that is likely not a false positive (but albeit a minor problem that doesn't effect our tests) is:

```
lib/pkg_editor/test/pkg_editor_test.cpp:336:3:
  Type: Copy-paste error (COPY_PASTE_ERROR)

lib/pkg_editor/test/pkg_editor_test.cpp:317:3:
  original: "strlen(file0)" looks like the original copy.
lib/pkg_editor/test/pkg_editor_test.cpp:336:3:
  copy_paste_error: "file0" in "strlen(file0)" looks like a copy-paste error.
lib/pkg_editor/test/pkg_editor_test.cpp:336:3:
  remediation: Should it say "file1" instead?
```
Taking a look at the code for that, it does look like it should be `strlen(file1)` instead of `strlen(file0)`, but it doesn't matter since `strlen(file0) == strlen(file1)`. I'll make a PR for this in the future when I finish with Coverity issues for our `test` dir.

IlanTruanovsky:
@pcolberg I've removed the comment. This should still make Coverity happy.

PR #263, opened 2023-01-31
Fix Coverity issues for `acl_mem_test.cpp`

IlanTruanovsky:
This mainly fixes various PRINTF_ARGS errors.

### DIFF
--- a/test/acl_mem_test.cpp
+++ b/test/acl_mem_test.cpp
@@ -1502,7 +1502,7 @@ MT_TEST(acl_mem, read_write_buf) {
                 int(flagstrial2));
 
         ACL_LOCKED(
-            acl_print_debug_msg("\n\nBEGIN trial %d flagstrial (%d,%d)\n",
+            acl_print_debug_msg("\n\nBEGIN trial %zu flagstrial (%zu,%zu)\n",
                                 trial, flagstrial1, flagstrial2));
         ACL_LOCKED(acl_print_debug_msg(" result0 is '%s'\n", resultbuf));
 
@@ -1580,7 +1580,8 @@ MT_TEST(acl_mem, read_write_buf) {
             // If COPY_HOST_PTR is specified, then the data is already
             // in mem1, as part of create-buffer operation.
             ACL_LOCKED(acl_print_debug_msg(
-                " result7 data '%s'\n", mem1->block_allocation->range.begin));
+                " result7 data '%s'\n",
+                (char *)mem1->block_allocation->range.begin));
           }
           ACL_LOCKED(acl_print_debug_msg(" result2 is '%s'\n", resultbuf));
           ACL_LOCKED(acl_print_debug_msg("Enqueue read\n"));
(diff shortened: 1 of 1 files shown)


IlanTruanovsky:
There is an additional Coverity error for this file that is causing me some confusion and that I will need to look into some more before implementing. This issue should be added to the PR before being merged with main. This error output is quite large, so I'll only leave in the important bits so we can discuss it:

```
test/acl_mem_test.cpp:2655:5:
  Type: Out-of-bounds access (ARRAY_VS_SINGLETON)

...

test/acl_mem_test.cpp:2655:5:
  11. address_of: Taking address with "&fill_event" yields a singleton pointer.
...

src/acl_event.cpp:988:3:
  13.1.18.33.4.2. path: Condition "num_events > 0", taking true branch.
src/acl_event.cpp:988:3:
  12.1.18.33.4.3. path: Condition "events == NULL", taking false branch.
src/acl_event.cpp:993:3:
  12.1.18.33.4.4. path: Condition "num_events == 0", taking true branch.
src/acl_event.cpp:993:3:
  12.1.18.33.4.5. path: Condition "events != NULL", taking false branch.
src/acl_event.cpp:999:3:
  12.1.18.33.4.6. path: Condition "i < num_events", taking true branch.
src/acl_event.cpp:1000:5:
  12.1.18.33.4.7. ptr_arith: Performing pointer arithmetic on "events" in expression "events + i".
```

To explain what's happening, we pass in a pointer to a single event `fill_event` to a function that at first glance seems to increment it by `i`, making it point to undesired memory if `i > 1`. However, this never happens in practice since we pass in the number of events and `i < # of events`.

What's even stranger is that Coverity is contradicting itself in its own analysis... 
According to the output, It's assuming `num_events > 0` and that `num_events == 0`, and a similar contradiction happens with `events`. Not sure what's up with that.

IlanTruanovsky:
The easiest (and maybe only?) solution to this would be to mark it as a false positive using a comment. I don't think it's possible to remove this Coverity issue by rearranging/reordering nearby code.

IlanTruanovsky:
> The Coverity issue is resolved by passing an array:
> 
> ```diff
> --- a/test/acl_mem_test.cpp
> +++ b/test/acl_mem_test.cpp
> @@ -2652,9 +2652,10 @@ MT_TEST(acl_mem, fill_image) {
>      CHECK_EQUAL(CL_INVALID_EVENT_WAIT_LIST,
>                  clEnqueueFillImage(m_cq, image, fill_color_int, origin, region,
>                                     1, NULL, &fill_event)); // invalid wait list
> +    cl_event event_wait_list[1];
>      CHECK_EQUAL(CL_INVALID_EVENT_WAIT_LIST,
>                  clEnqueueFillImage(m_cq, image, fill_color_int, origin, region,
> -                                   0, &fill_event,
> +                                   0, event_wait_list,
>                                     &fill_event)); // invalid wait list
>  
>      CHECK_EQUAL(1, acl_ref_count(image));
> ```
> 
> I have yet to understand what the underlying issue is.

Yesterday I was attempting this solution but I didn't finish running the Coverity scan.

Glad to know it works!

Issue #245, opened 2023-01-13
Resolve Coverity static analysis issues


pcolberg:
The runtime project is transitioning from Klocwork to Coverity for static code analysis. This issue tracks the number of remaining Coverity issues on a semi-regular basis. The initial goal is to reduce the number of Coverity issues to zero and fail the [Coverity workflow](https://github.com/intel/fpga-runtime-for-opencl/blob/dd7c2cd1154c4dcfbef62852abb3d5fa30707bf9/.github/workflows/coverity.yml#L68) with a non-zero exit status on new issues. So far all Coverity issues could be addressed through reasonable code changes that fix potential bugs and/or improve code quality.

When updating this issue, attach a copy of the zip file from the Coverity workflow artifacts for archival, since artifacts are purged after some time.

To count issues, take the number of blank lines and divide by 2:
```
echo "scale=0; $(grep -c '^$' cov-errors.txt) / 2" | bc -l
```

IlanTruanovsky:
> To count issues, take the number of paragraphs and divide by 2:

The `cov-analyze` command that runs in the workflow provides detailed information regarding the number of defects. 

I.e. for the first Coverity run: https://github.com/intel/fpga-runtime-for-opencl/actions/runs/3525211581/jobs/5913259883#step:9:53

Should we be using that instead?

zibaiwan:
Thanks everyone's effort, especially @tylerzhao7684, we have 0 Coverity issue now

PR #266, opened 2023-02-02
workflows: exclude CppUTest from Coverity static analysis

pcolberg:
This is third-party code that is not shipped to customers.

https://github.com/intel/fpga-runtime-for-opencl/issues/245#issuecomment-1382399646
https://github.com/intel/fpga-runtime-for-opencl/issues/245#issuecomment-1407201339

Cc @IlanTruanovsky


zibaiwan (approved):
Thanks Peter!

PR #268, opened 2023-02-03
Reorganize docs directory

sophimao:
Propose new directory structure for docs:
```
docs
└── features
    └── eager_progess.md (to be uploaded)
    └── ...
└── images
    └── <images used in mds under features/opencl>
└── opencl
    └── buffers.md
    └── ...
```


sophimao:
> Having a subdirectory for the images seems fine, but I am a bit unsure about introducing a substructure for the documents this early, for merely two files. Then, the question what directory names to pick is quite difficult if the number of documents is so small; e.g., you could consider both of the example documents "features". What do you think?

Right I agree, do you think if it would be good if I just keep the original directory structure, but just add a directory for image for now (in that case I'll just close this PR)?

sophimao:
Will retain original directory structure for now

PR #269, opened 2023-02-03
Device global copy kernel implementation

intel-liudean:
Migrated from Sherry's PR at https://github.com/intel/fpga-runtime-for-opencl/pull/65.

TODO: 
1) figure out what to do with acl_auto_configure.cpp and acl_auto_configure_test.cpp
2) address the problem "device global pointer is being mocked by passing an actual device allocation into the function"
3) Proof read and properly test the code


intel-liudean:
Copy Kernel is no longer needed to support the 2023.2 deliverable device_globals with host _access.
This PR will NOT be checked in.

PR #277, opened 2023-02-06
Fix `TAINTED_SCALAR` Coverity issues for `pkg_editor.c`

IlanTruanovsky:
Several values that were being read through files were not being checked to see if they meet certain requirements.

For one of the issues, the `info.name_length` variable was not being checked to see if it was less than the size of `name` when passed into `read_data`. This was a simple fix.

For the other `TAINTED_SCALAR` issue, the `info.file_length` value was never checked. Since this was being passed into a `malloc` call, it would be possible for the malloc call to use too much memory if the `info.file_length` number was purposefully set to be large. There were two possible solutions to this that I could think of
1. Add an upper bound to the size of `info.file_length`. This maintains much of the same code, but also increases the restrictions of the data being passed into the `acl_pkg_unpack_buffer_or_file` function.
2. Do not malloc an array of length `info.file_length` and instead reuse a constant sized array. This is the solution I implemented. 

Instead of mallocing and allocating memory, I added a while loop to reuse a constant sized array for multiple iterations. The only problem with this that I can foresee is it taking too much time and having multiple iterations of the while loop. However, when I ran the unit tests with and without this change, it took the same amount of time (although I'm not sure if the unit tests cover this case thoroughly). Are there any issues with this that come to mind, @pcolberg?

The error output is large, so I'll trim it down to just the issue without the code trace.

Fixes:
```
lib/pkg_editor/src/pkg_editor.c:1632:5:
  Type: Untrusted value as argument (TAINTED_SCALAR)

lib/pkg_editor/src/pkg_editor.c:1681:11:
  Type: Untrusted allocation size (TAINTED_SCALAR)
```

### DIFF
--- a/lib/pkg_editor/test/pkg_editor_test.cpp
+++ b/lib/pkg_editor/test/pkg_editor_test.cpp
@@ -53,6 +53,8 @@ using random_bytes_engine =
                                  unsigned int>;
 namespace fs = std::experimental::filesystem::v1;
 
+int tmpCount = 0;
+
 static void l_remove_file(const char *filename) {
 #ifdef _WIN32
   _unlink(filename);
@@ -457,8 +459,9 @@ static bool is_same_tmpdir(const std::vector<fs::path> &files,
 
 TEST(package, unpack) {
   int result;
-  fs::path tmpdir = "tmp";
-  std::string tmpdir_string = tmpdir.string();
+  std::string tmpdir_string = "tmp" + std::to_string(tmpCount);
+  tmpCount++;
+  fs::path tmpdir = tmpdir_string;
   const char *tmpdir_c_str = tmpdir_string.c_str();
   std::vector<fs::path> files = generate_tmp_folder(tmpdir);
 
@@ -473,12 +476,14 @@ TEST(package, unpack) {
 
   // Compare some files to be sure that they are the same.
   CHECK(is_same_tmpdir(files, PACK_UNPACK_DIR));
+  system(("rm -rf " + tmpdir_string).c_str());
 }
 
 TEST(package, unpack_buffer) {
   int result;
-  fs::path tmpdir = "tmp";
-  std::string tmpdir_string = tmpdir.string();
+  std::string tmpdir_string = "tmp" + std::to_string(tmpCount);
+  tmpCount++;
+  fs::path tmpdir = tmpdir_string;
   const char *tmpdir_c_str = tmpdir_string.c_str();
   std::vector<fs::path> files = generate_tmp_folder(tmpdir);
   // Create a known good input.
@@ -509,12 +514,14 @@ TEST(package, unpack_buffer) {
   CHECK_EQUAL(true, files_same("test/pkg_editor_test.cpp",
                                PACK_UNPACK_DIR "/test/pkg_editor_test.cpp"));
   CHECK(is_same_tmpdir(files, PACK_UNPACK_DIR));
+  system(("rm -rf " + tmpdir_string).c_str());
 }
 
 TEST(package, unpack_buffer_stdin) {
   int result;
-  fs::path tmpdir = "tmp";
-  std::string tmpdir_string = tmpdir.string();
+  std::string tmpdir_string = "tmp" + std::to_string(tmpCount);
+  tmpCount++;
+  fs::path tmpdir = tmpdir_string;
   const char *tmpdir_c_str = tmpdir_string.c_str();
   std::vector<fs::path> files = generate_tmp_folder(tmpdir);
 
@@ -563,5 +570,6 @@ TEST(package, unpack_buffer_stdin) {
   CHECK_EQUAL(true, files_same("test/pkg_editor_test.cpp",
                                PACK_UNPACK_DIR "/test/pkg_editor_test.cpp"));
   CHECK(is_same_tmpdir(files, PACK_UNPACK_DIR));
+  system(("rm -rf " + tmpdir_string).c_str());
 }
 #endif
--- a/lib/pkg_editor/test/pkg_editor_test.cpp
+++ b/lib/pkg_editor/test/pkg_editor_test.cpp
@@ -12,9 +12,17 @@
 #pragma warning(pop)
 #endif
 
+// Replace this library with <filesystem> once we move to GCC 8 or newer
+// versions
+#define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING 1
+#include <experimental/filesystem>
+
+#include <climits>
+#include <random>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <string>
 
 #include "pkg_editor/pkg_editor.h"
 #include <assert.h>
@@ -39,6 +47,11 @@
 #define SAMPLE_FILE ".sample_file.elf"
 #define PACK_UNPACK_FILE ".pack_unpack"
 #define PACK_UNPACK_DIR ".pack_unpack_dir"
+
+using random_bytes_engine =
+    std::independent_bits_engine<std::default_random_engine, CHAR_BIT,
+                                 unsigned int>;
+namespace fs = std::experimental::filesystem::v1;
 
 static void l_remove_file(const char *filename) {
 #ifdef _WIN32
@@ -379,7 +392,7 @@ TEST(package, pack) {
   CHECK_EQUAL(1, result);
 }
 
-static bool files_same(const char *f1, const char *f2) {
+static bool files_same(const fs::path f1, const fs::path f2) {
   std::ifstream file1(f1, std::ifstream::ate | std::ifstream::binary);
   std::ifstream file2(f2, std::ifstream::ate | std::ifstream::binary);
   file1.exceptions(std::ifstream::failbit | std::ifstream::badbit);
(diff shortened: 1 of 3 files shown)


IlanTruanovsky:
> Running the unit tests, how often is that inner loop executed in a single call to `read_data()` using your chunk size of 64 kB? How does it change if you decrease the chunk size?

I did some testing, seems like the loop in `read_data()` only ever does 1 iteration regardless of the size of the buffer (even at 1 kB).

IlanTruanovsky:
I tried replicating your experiment, but mine passes.

I think you're inserting the 1GB and 100MB files in the wrong place, it should be located under `<your build dir>/lib/pkg_editor/test/test/test.100M` (notice the double `/test` directories)

IlanTruanovsky:
@pcolberg thanks for all your debugging! I'll take a look at what needs to be done and commit the necessary changes.

IlanTruanovsky:
> So far we have found the following gaps in the `pkg_editor` test suite:
> 
> * [`files_same`](https://github.com/intel/fpga-runtime-for-opencl/blob/f66baab8b65ad9856653661f651f71d8da62fd9c/lib/pkg_editor/test/pkg_editor_test.cpp#L382-L397) does not throw an exception when test files are absent.
> * Zero-length uncompressed files are not tested.

Funnily enough, [this test](https://github.com/intel/fpga-runtime-for-opencl/blob/main/lib/pkg_editor/test/pkg_editor_test.cpp#L412-L413) fails when I add a check to see if the file exists. I need to investigate this a bit more.

IlanTruanovsky:
> > So far we have found the following gaps in the `pkg_editor` test suite:
> > 
> > * [`files_same`](https://github.com/intel/fpga-runtime-for-opencl/blob/f66baab8b65ad9856653661f651f71d8da62fd9c/lib/pkg_editor/test/pkg_editor_test.cpp#L382-L397) does not throw an exception when test files are absent.
> > * Zero-length uncompressed files are not tested.
> 
> Funnily enough, [this check](https://github.com/intel/fpga-runtime-for-opencl/blob/main/lib/pkg_editor/test/pkg_editor_test.cpp#L412-L413) fails when I add a check to see if the file exists. I need to investigate this a bit more.

It seems like CMakeLists.txt was never packed to begin with, so it doesn't exist. It also doesn't exist in the cwd while the unit test is running. I'll go ahead and replace the file name with something that exists.

IlanTruanovsky:
@pcolberg I added better unit testing for the unpacking part of the pkg_editor unit tests, including:
- random file generation and checking
- large file generation and checking (10 MB)
- empty file generation and checking

Your review is much appreciated!

lib/pkg_editor/test/pkg_editor_test.cpp
```diff
@@ -40,13 +43,16 @@
...
+
 static void l_remove_file(const char *filename) {
 #ifdef _WIN32
   _unlink(filename);
 #else
   unlink(filename);
 #endif
-}
+} // namespace
+  // fs=std::experimental::filesystem::v1staticvoidl_remove_file(constchar*filename)
```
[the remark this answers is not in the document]
IlanTruanovsky: I'm not sure how that happened... I must have been fooling around with the line with `namespace fs = std::experimental::filesystem::v1;` and somehow must have commented that out.

I'll remove it.

lib/pkg_editor/test/pkg_editor_test.cpp
```diff
@@ -396,11 +404,57 @@ static bool files_same(const char *f1, const char *f2) {
...
+
+  generated_files[num_files - 2] = "./tmp/large_file";
+  generate_random_file(generated_files[num_files - 2], 10000000); // ~10MB
+
+  num_files -= 2;
+
+  for (int i = 0; i < num_files; i++) {
+    char file_template[] = "./tmp/fileXXXXXX";
+    // We don't care about the file descriptor that is returned
+    close(mkstemp(file_template));
```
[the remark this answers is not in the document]
IlanTruanovsky: Ah, of course! It seems I was overengineering things.

IlanTruanovsky:
@pcolberg I've made all the changes you requested. You can review the code once again. If everything looks good, we can try running the integration test again. However, it does look like the [Windows build is failing](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4286895476/jobs/7467063949), so I'll look into that.

I also ran the unit tests before the TAINTED_SCALAR changes as you suggested, and it does pass.

IlanTruanovsky:
Hmm, when I try to compile it on Windows myself, I get:
```
...\include\random(2401): error C2338: invalid template argument for uniform_int_distribution: N4659 29.6.1.1 [rand.req.genl]/1e requires one of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long
```
and
```
...\include\random(2401): error C2338: note: char, signed char, unsigned char, int8_t, and uint8_t are not allowed
```

IlanTruanovsky:
> Hmm, when I try to compile it on Windows myself, I get:
> 
> ```
> ...\include\random(2401): error C2338: invalid template argument for uniform_int_distribution: N4659 29.6.1.1 [rand.req.genl]/1e requires one of short, int, long, long long, unsigned short, unsigned int, unsigned long, or unsigned long long
> ```
> 
> and
> 
> ```
> ...\include\random(2401): error C2338: note: char, signed char, unsigned char, int8_t, and uint8_t are not allowed
> ```

I fixed this on Windows by removing all the `<char>` from the uniform distributions that have them. However, I get the following error: 

```
LINK : fatal error LNK1181: cannot open input file 'stdc++fs.lib'
```

This is really strange since this seems completely unrelated...

IlanTruanovsky:
@pcolberg It's been a while, but I was able to get this PR to pass, though there was one complication that I'd like your input on.

I'm using [this flow](https://github.com/intel/fpga-runtime-for-opencl/pull/277/files#diff-4175dc63787a6fec2170d67a1477ebaa7b5c0a2c83a559a0bc532ab60fc71ff4R430-R435) to remove the tmp dir when creating the random files during our testing. This code does pass CI, but it did not work when I tested locally on a Windows machine (it errors out with `Exit code 0xc0000409`). The same happened when I replaced this code with `fs::remove_all(tmpdir)`, which [also failed Windows CI ](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4887555329/jobs/8724268644) (not 100% sure this is the correct workflow run). In both cases, Linux worked fine.

So what should be done about this? We can either:
- Leave it as is now as it passes CI. It could be the case that it did not pass locally due to my Windows machine just being slow and causing errors.
- Remove this part of the code completely and document the function's new behavior. The tests still pass anyways, and we regenerate the files every time we call the function.
- Investigate other options for removing the folder. I'm not sure what other options there are, but maybe you have ideas?

I think option #2 is the best, but I'd like to know what you think.

zibaiwan:
I had an offline discussion with Ilan. Ideally we don't want to create tmp files in the tmp directory during the unit testing without cleaning them up. @tylerzhao7684 , can you please investigate the error code that Ilan posted in the earlier comment to see what that means on the windows machine and how we can avoid it?

We definitely don't want to have intermittent build failures.

lib/pkg_editor/test/pkg_editor_test.cpp
```diff
@@ -396,11 +411,59 @@ static bool files_same(const char *f1, const char *f2) {
...
+// Creates a folder consisting of a certain number of files with random contents
+// and random sizes
+static std::vector<fs::path> generate_tmp_folder(const fs::path &tmpdir) {
+  std::vector<fs::path> files;
+
+  std::mt19937 gen{};
+  std::uniform_int_distribution dis{0, 100000};
+
+  int system_result = 0;
+  std::string remove_command = "rm -rf " + tmpdir.string();
```
tylerzhao7684, line 431:
```suggestion
  std::string remove_command = "cd " + tmpdir.string() + " && rm -rf *";
```

tylerzhao7684: Seems like this change will properly delete the tmp directory after some local testing

tylerzhao7684 (changes requested):

lib/pkg_editor/test/pkg_editor_test.cpp
tylerzhao7684: The previous error's root cause is the tmp directory not properly removed (and caused some corruption on windows specifically?)

IlanTruanovsky: This seems to work, though I have no idea why :smile:
If the tmp directory is corrupt, why would "cd"ing into it fix it...

Anyways, I would also add a check to see if `fs::exists(tmpdir)` is true before doing the system call to avoid any other error caused by "cd"ing into the directory.

Can you please make this change and update the PR @tylerzhao7684? Thanks!

tylerzhao7684: Seems like my approach still didn't solve the root problem
In my suggested code I didn't remove the tmp directory and it did not clean up after the unit tests finished
However, removing the tmp directory will throw and win32 exception, still trying to figure out why...

tylerzhao7684: Still not sure what is going on...
When trying to delete and recreate the tmp directory in the second call of `generate_tmp_folder` will fail (The tmp directory will not be created in the second call)
Exception is thrown in `generate_random_file` because the `tmp` directory did not get created in the second call but the program tries to put files inside the non-existent `tmp` directory.

Tried debugging this: 
During the second call of `generate_tmp_folder`, `system("ls")` was used to see if there is any remaining `tmp` directory after `rm -rf tmp`, there was no `tmp` directory
However, `system("ls")` was used again after `fs::create_directory(tmpdir)`, yet there is still no `tmp` directory
I replaced `fs::create_directory(tmpdir)` with `mkdir tmp` and there was this error message:
`A subdirectory or file tmp already exists` even though `ls` did not show any `tmp` directory or file

However there is a workaround, which is to name the `tmp` directory differently in each unit test. This avoids removing and re-creating the directory that caused this issue
Tested this workaround locally and the `tmp` directories are properly cleaned up
Will update the PR with the workaround soon...

tylerzhao7684 (approved):
Windows unit tests now passing

zibaiwan:
Can you please do an integration testing as well? Thanks!

tylerzhao7684:
> Can you please do an integration testing as well? Thanks!

Integration test passed https://github.com/intel-innersource/applications.fpga.oneapi.stella/pull/6976

tylerzhao7684:
> Can you please do an integration testing as well? Thanks!

It passed

zibaiwan (approved):
Thanks Tyler. The change looks good to me. Just need a few hw tests to pass and I will merge.

zibaiwan:
Thanks Tyler!

PR #278, opened 2023-02-15
Auto discovery change for device global

intel-liudean:
For each device global, the init_mode field is changed to can_skip_programming field, a new field reset_on_reuse is added at the end.

### DIFF
--- a/include/acl.h
+++ b/include/acl.h
@@ -512,25 +512,16 @@ typedef enum {
   ACL_DEVICE_GLOBAL_HOST_ACCESS_TYPE_COUNT
 } acl_device_global_host_access_t;
 
-// Enum values here also need to match the SPIRV spec for device
-// global in the above link for acl_device_global_host_access_t.
-// ACL_DEVICE_GLOBAL_INIT_MODE_TYPE_COUNT is used for validation in
-// autodiscovery string parsing and should remain the last constant
-// in the enum.
-typedef enum {
-  ACL_DEVICE_GLOBAL_INIT_MODE_REPROGRAM,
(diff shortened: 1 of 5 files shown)


zibaiwan:
@intel-liudean , thanks for working on this!

I haven't finished the review yet but I think you need to make more updates in the acl_auto_configure_test.cpp. Take a look at the Sophie's change earlier: https://github.com/intel/fpga-runtime-for-opencl/commit/6f68280f5ba865bde96d27ea63f82b411ee3958a, there are many hard-coded device global in several tests, all of them require fields update I believe.

Also please squash those two commits into 1 single commit and provide the description in the commit message and in the PR title. Thanks!

include/acl.h
```diff
@@ -514,23 +514,17 @@ typedef enum {
 
 // Enum values here also need to match the SPIRV spec for device
```
sophimao, line 515: Since the enum was removed, I think you can also remove this part of the comment.

intel-liudean:
> @intel-liudean , thanks for working on this!
> 
> I haven't finished the review yet but I think you need to make more updates in the acl_auto_configure_test.cpp. Take a look at the Sophie's change earlier: [6f68280](https://github.com/intel/fpga-runtime-for-opencl/commit/6f68280f5ba865bde96d27ea63f82b411ee3958a), there are many hard-coded device global in several tests, all of them require fields update I believe.
> 
> Also please squash those two commits into 1 single commit and provide the description in the commit message and in the PR title. Thanks!

I have addressed your feedback @zibaiwan. Please take a look again.

zibaiwan (approved):
Looks good to me! Thanks!

PR #280, opened 2023-02-17
acl_hal_mmd: declare dlopen() wrappers as static functions

pcolberg:
These functions are only used within the translation unit.


zibaiwan (approved):
Thanks @pcolberg !

PR #284, opened 2023-04-04
Program scope hostpipe support

zibaiwan:
1. Support program scope hostpipe, CSR pipe and non-CSR pipe.
2. Register program scoped hostpipes during program creation.
3. Create new command op and device op for the hostpipe read and write
4. Unit test is only for the auto-discovery string. Other unit tests will be added through other PR.
5. Most of the Klocwork issues are the existing issues but listed as the new issues due to I modified the source code for the test, but I will attempt to fix all of them.
6. More error checking needs to be done. Will add through a new PR.

### DIFF
--- a/src/acl_hostch.cpp
+++ b/src/acl_hostch.cpp
@@ -14,8 +14,11 @@
 // Internal headers.
 #include <acl.h>
 #include <acl_context.h>
+#include <acl_device_op.h>
+#include <acl_event.h>
 #include <acl_hostch.h>
 #include <acl_mem.h>
+#include <acl_platform.h>
 #include <acl_util.h>
 
 #ifdef __GNUC__
@@ -24,7 +27,7 @@
 
 /* Local Functions */
 static cl_int l_push_packet(unsigned int physical_device_id, int channel_handle,
-                            void *host_buffer, size_t write_size) {
+                            const void *host_buffer, size_t write_size) {
   size_t pushed_data;
   int status = 0;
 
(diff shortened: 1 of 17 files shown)


zibaiwan:
@sophimao , while I am still working on the Klocwork fixes, please take a look at the change. Thanks!

src/acl_hostch.cpp
```diff
@@ -682,6 +685,399 @@ clUnmapHostPipeIntelFPGA(cl_mem pipe, void *mapped_ptr, size_t size_to_unmap,
...
+      acl_set_device_op_execution_status(op, -1);
+      return;
+    }
+
+    data_reg = static_cast<uintptr_t>(parsed);
+    ready_reg = static_cast<uintptr_t>(
+        parsed + 8); // ready reg is data reg shift by 8 byte, move this to the
+                     // autodiscovery string maybe
+    valid_reg = static_cast<uintptr_t>(
+        parsed + 16); // valid reg is ready reg shift by 8 byte
```
sophimao, line 728: Perhaps instead of 8 and 16 we can set this as preprocessor variable or `static constexpr unsigned`? Though It's not that hard to grasp from the code that this these numbers are offsets so probably also okay to leave as it.

src/acl_hostch.cpp
```diff
@@ -682,6 +685,399 @@ clUnmapHostPipeIntelFPGA(cl_mem pipe, void *mapped_ptr, size_t size_to_unmap,
...
+    size_t pushed_data;
+    try {
+      parsed = std::stoull(host_pipe_info.csr_address, nullptr);
+    } catch (const std::exception &) {
+      acl_set_device_op_execution_status(op, -1);
+      return;
+    }
+    data_reg = static_cast<uintptr_t>(parsed);
+    valid_reg = static_cast<uintptr_t>(
+        parsed + 8); // valid reg is data reg shift by 8 byte, move this to the
```
sophimao, line 815: Note the 8 here as well

sophimao (approved):
Thanks Zibai! Overall looks good to me, just 2 minor comments as listed below.

zibaiwan:
@sophimao , thanks for your review. I have addressed your comments, squashed the commits. Please take a look when you have a chance!

PR #285, opened 2023-04-04
Fix kernel id - CSR address mapping issue

sophimao:
Currently in simulation we need to match CSR address with kernel id in order to know which start register the runtime is writing into and set the corresponding start signal in the simulation control BFM. This change sends the kernel id CSR address mapping information to the simulation runtime through a new simulation MMD function.

### DIFF
--- a/include/acl_hal_mmd.h
+++ b/include/acl_hal_mmd.h
@@ -149,6 +149,12 @@ typedef struct {
   // accounted and returned in a subsequent invocation of this function.
   void (*aocl_mmd_simulation_streaming_kernel_done)(
       int handle, const std::string &signal_name, unsigned int &finish_counter);
+
+  // Pass kernel-id to csr-address mapping read from the current binary
+  // to the simulation runtime, so that it can detect which start register
+  // has been written by the runtime.
+  void (*aocl_mmd_simulation_set_kernel_cra_address_map)(
+      int handle, const std::vector<uintptr_t> &kernel_csr_address_map);
 } acl_mmd_dispatch_t;
 
 typedef struct {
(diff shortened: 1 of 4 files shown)


sophimao:
Hi @zibaiwan , this is the change to runtime I mentioned before, would you mind taking a look? Thanks!

include/acl_hal_mmd.h
```diff
@@ -149,6 +149,9 @@ typedef struct {
   // accounted and returned in a subsequent invocation of this function.
   void (*aocl_mmd_simulation_streaming_kernel_done)(
       int handle, const std::string &signal_name, unsigned int &finish_counter);
+
```
zibaiwan, line 152: Hi Sophie, can you please add some comments to describe this mmd function? Thanks!

zibaiwan (dismissed):
Hi Sophie, looks good to me except a minor issue.

src/acl_hal_mmd.cpp
```diff
@@ -2854,3 +2859,11 @@ void acl_hal_mmd_simulation_streaming_kernel_done(
       .mmd_dispatch->aocl_mmd_simulation_streaming_kernel_done(
           device_info[physical_device_id].handle, kernel_name, finish_counter);
 }
+
+void acl_hal_mmd_simulation_set_kernel_cra_address_map(
+    unsigned int physical_device_id,
+    const std::vector<uintptr_t> kernel_csr_address_map) {
```
zibaiwan, line 2865: One more nitpick, I think it's better to pass kernel_csr_address_map by reference.

zibaiwan (dismissed):
Thanks Sophie!

sophimao:
Hi @zibaiwan , sorry but looks like updating commit message invalidates your review, would you mind approving again?

zibaiwan (approved):

PR #286, opened 2023-04-12
Modify USM device allocation information returned by clGetDeviceInfo

sophimao:
According to [USM extension spec](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_unified_shared_memory.html#_device_allocations), clDeviceMemAllocINTEL will return `NULL` and `CL_INVALID_OPERATION` status if the device has property `CL_DEVICE_DEVICE_MEM_CAPABILITIES_INTEL` equal to 0. Before, the runtime always return 1 for device allocation capabilities as all legacy devices support device allocations. This is no longer true in IP Authoring flow.


zibaiwan (approved):
Thanks Sophie for fixing this!

PR #288, opened 2023-04-17
Skip non-program scoped hostpipe registration in the program build.

zibaiwan:
This fixes a regression introduced by #284 in the opencl hw flow using the old deprecated opencl kernel hostpipe.


sophimao (approved):
Thanks Zibai!

PR #289, opened 2023-04-26
Fix CSR Pipe

zibaiwan:
The previous return code checking was wrong. Also skip some register writing in the non-blocking case.

### DIFF
--- a/src/acl_hostch.cpp
+++ b/src/acl_hostch.cpp
@@ -736,17 +736,30 @@ void acl_read_program_hostpipe(void *user_data, acl_device_op_t *op) {
 
     // start the CSR read
 
-    // Checking if the data is valid, blocking
+    // If Blocking, wait until the data is valid.
+    // If Non-blocking, just read once and report failure if not valid.
     do {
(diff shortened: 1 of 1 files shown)


src/acl_hostch.cpp
```diff
@@ -741,12 +741,24 @@ void acl_read_program_hostpipe(void *user_data, acl_device_op_t *op) {
       acl_get_hal()->read_csr(host_pipe_info.m_physical_device_id, valid_reg,
                               (void *)valid_value_pointer,
                               (size_t)sizeof(uintptr_t));
-    } while (valid_value != 1);
+    } while (blocking && valid_value != 1);
```
sophimao, line 744: Is the logic here that if non-blocking, read csr valid register once, otherwise read csr valid register until valid is 1? If that is the case can you modify the comment on line 739 to say if non-blocking only read once? The comment only mentions blocking so I was a bit confused at first.

zibaiwan: Thanks Sophie. I have addressed the comment!

PR #290, opened 2023-05-01
Coverity Fix: Close opened mmd libraries

tylerzhao7684:
`Type: Resource leak (RESOURCE_LEAK)`
`mmd_library` is not properly closed under several exit conditions, we need to close it before exiting

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -67,7 +67,7 @@ jobs:
       - run: cov-analyze --config config.xml --dir results --concurrency --security --rule --enable-constraint-fpp --enable-fnptr --enable-virtual
       - run: cov-format-errors --text-output-style multiline --dir results --filesort --file "$PWD" --strip-path "$PWD" > cov-errors.txt
       - run: cat cov-errors.txt
-      - run: count=$(grep -c '^$' cov-errors.txt) && echo "$(( $count / 2 ))"
+      - run: count=$(grep -c '^$' cov-errors.txt) || true && echo "$(( $count / 2 ))"
       - uses: actions/upload-artifact@v3
         with:
           name: fpga-runtime-for-opencl-${{ github.sha }}-coverity-${{ github.run_id }}
--- a/src/acl_hal_mmd.cpp
+++ b/src/acl_hal_mmd.cpp
@@ -464,6 +464,24 @@ static void my_dlclose(void *library) {
 #endif
 }
 
+static void my_dlclose_no_assert(void *library) {
+#ifdef _WIN32
+  FreeLibrary((HMODULE)library);
+#else
+  dlclose(library);
+#endif
+}
+
+typedef struct my_dl_wrapper {
+  void *handle;
+  my_dl_wrapper(void *handle) { this->handle = handle; }
+  ~my_dl_wrapper() { my_dlclose_no_assert(this->handle); }
+  // prohibit copying to avoid double-close of handle
+  my_dl_wrapper(const my_dl_wrapper &) = delete;
+  my_dl_wrapper &operator=(const my_dl_wrapper &) = delete;
+} my_dl_wrapper;
+std::vector<std::unique_ptr<my_dl_wrapper>> mmd_libs;
+
 cl_bool l_load_board_functions(acl_mmd_dispatch_t *mmd_dispatch,
                                const char *library_name, void *mmd_library,
                                char *error_msg) {
@@ -632,12 +650,14 @@ cl_bool l_load_single_board_library(const char *library_name,
       if (result == CL_FALSE) {
         std::cout << "Error: Could not load board library " << library_name
                   << " due to failure to load symbols\n";
+        my_dlclose(mmd_library);
         return result;
       }
     }
     ++num_boards_found;
   }
 
+  mmd_libs.push_back(std::make_unique<my_dl_wrapper>(mmd_library));
   return CL_TRUE;
 }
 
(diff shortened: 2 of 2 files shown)


tylerzhao7684:
Seems like the mmd library is accessed globally, need another way to fix this

tylerzhao7684:
Confirmed that it is [this line](https://github.com/intel/fpga-runtime-for-opencl/pull/290/files#diff-19c39c37ac49dd54644d15ef56b292822bfd0b787aabe9bdadeed8154196aa6dR633) that causes the unit test to fail
@zibaiwan Any idea on this?
Coverity doesn't like mmd_library (A pointer) to be lost even though the library is accessed globally

tylerzhao7684:
I looked at https://github.com/intel/fpga-runtime-for-opencl/pull/227 and used a similar method to close the library
Also looked at https://github.com/intel/fpga-runtime-for-opencl/pull/264 and I think it caused a regression because in some cases `dlsym` need to be called on libraries that are opened globally
The current fix allows mmd library to be opened globally as well as automatically be closed when the runtime finishes (This fixes the Coverity case but not https://hsdes.intel.com/appstore/article/#/14018765067)
I'll run l3 to see if this also caused an regression

tylerzhao7684:
This fix is blocked until we remove pac_a10 and pac_s10 boards (At least) in test infra

tylerzhao7684:
@zibaiwan Integration test passed with this implementation
Review please!

zibaiwan (dismissed):

zibaiwan:
@tylerzhao7684 , as we are very close to merge this in. Assuming the window testing passes, could you please squash your 4 commits into 1 commit as they essentially all for fixing one issue?

The configuration in this repo is set to merge all 4 commits into the repo instead of squashing them into 1 automatically, to allow one large PR can have multiple meaningful commits. You can choose the way you like to do the squash, e.g., `rebease -i` or `reset --soft`, and then do a force push. Let me know if you have any questions! 

Thanks you!

zibaiwan:
@tylerzhao7684 , can you please take a look at the workflow failure? My guess is, `cov-errors.txt` is now empty based on the empty output of the `cat cov-errors.txt` because you fixed the last Coverity issue. However, I don't understand why `count=$(grep -c '^$' cov-errors.txt) && echo "$(( $count / 2 ))"` would fail in this case, instead of just outputting `0`.

One possible guess is: `cov-errors.txt` didn't exist and the `grep` failed, but it didn't make sense why `cat cov-errors.txt` didn't fail earlier. Can you please investigate a bit and this PR likely needs a supplement change in the workflow as well, otherwise the main branch workflow will show the failure status.

tylerzhao7684:
This previous workflow didn't fail because #277 was not merged in, so even if this PR fixes one coverity issue, the other one still remain as it is not merged
Here's the previous [workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4916197415/jobs/8779637472) 

Now it fails because it couldn't find cov-errors.txt, which is a good sign because now we don't have coverity issues anymore!
I'll work on a small fix in this PR to make this workflow succeed when cov-errors.txt is not found

zibaiwan:
> This previous workflow didn't fail because #277 was not merged in, so even if this PR fixes one coverity issue, the other one still remain as it is not merged Here's the previous [workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4916197415/jobs/8779637472)
> 
> Now it fails because it couldn't find cov-errors.txt, which is a good sign because now we don't have coverity issues anymore! I'll work on a small fix in this PR to make this workflow succeed when cov-errors.txt is not found

I am not 100% sure if it indeed failed due to cov-errors.txt is not found, because in the workflow, `cat cov-errors.txt` step didn't fail. I tested it locally on one of the linux machine, this would give an error like this `cat: dummy.txt: No such file or directory`, but maybe the workflow machine has a different linux distribution therefore no error? But if it indeed caused by no file found, I 100% agree this is a good sign!

zibaiwan:
> This previous workflow didn't fail because #277 was not merged in, so even if this PR fixes one coverity issue, the other one still remain as it is not merged Here's the previous [workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4916197415/jobs/8779637472)
> 
> Now it fails because it couldn't find cov-errors.txt, which is a good sign because now we don't have coverity issues anymore! I'll work on a small fix in this PR to make this workflow succeed when cov-errors.txt is not found

A separate commit for the workflow fix is more appropriate (no need to squash) but can be merged in the same PR.

tylerzhao7684:
> > This previous workflow didn't fail because #277 was not merged in, so even if this PR fixes one coverity issue, the other one still remain as it is not merged Here's the previous [workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/4916197415/jobs/8779637472)
> > Now it fails because it couldn't find cov-errors.txt, which is a good sign because now we don't have coverity issues anymore! I'll work on a small fix in this PR to make this workflow succeed when cov-errors.txt is not found
> 
> I am not 100% sure if it indeed failed due to cov-errors.txt is not found, because in the workflow, `cat cov-errors.txt` step didn't fail. I tested it locally on one of the linux machine, this would give an error like this `cat: dummy.txt: No such file or directory`, but maybe the workflow machine has a different linux distribution therefore no error? But if it indeed caused by no file found, I 100% agree this is a good sign!

You are correct, there is a cov-errors.txt. 
The workflow failed because grep returns non-zero exit code when it finds no match, I added the fix now

zibaiwan (approved):

PR #292, opened 2023-05-10
Coverity workflow fail if issue exist

tylerzhao7684:
Currently testing...


tylerzhao7684:
This workflow now works as follows:
- If a file change in this PR introduces a Coverity issue & that file originally does not have any Coverity issues, then this workflow would fail, to notify that there are Coverity issues
- If a file change in this PR edits a file that currently has Coverity issues (Not introduced by the current PR), then this workflow would fail. Error message would tell users to review Coverity issues manually and add a **coverity-override** tag to this PR when it is confirmed that the current PR does not introduce and new Coverity issues
- If **coverity-override** tag is added to the PR, then Coverity workflow is skipped

The reason why we cannot easily detect if the current PR introduces new Coverity issues when it modified an existing file that has Coverity issues, is because we only knows the Coverity issue by examining the output of `cov-format-errors --text-output-style multiline --dir results --filesort --file "$(realpath ..)" --strip-path "$(realpath ..)"` (i.e. cov-error-base.txt or cov-error.txt). A simple diff is unable to catch if a new Coverity issue is introduced because line number where the Coverity issues are located could be different

tylerzhao7684:
Successful [Coverity workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5522700574/jobs/10072653479?pr=292) when there is no Coverity Issues

tylerzhao7684:
This workflow is [failing](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5523189327/jobs/10073766571?pr=292) because it found a Coverity issue due to [this change](https://github.com/intel/fpga-runtime-for-opencl/pull/292/commits/9ea8060f2399dba74cdeba05b409a96afa31a69b)

tylerzhao7684:
The key reason why there are 2 builds in the coverity-pull-request workflow is because it "ignores" the existing coverity case if you don't edit the same file where there is an existing coverity case. It will be ignored because that case will show up in both cov-errors.txt and cov-errors-base.txt, and it will be ignored when we diff the two files. 

The only case we need the coverity-override label is when we actually edit the file that has an existing coverity case in this PR. When that happens, the diffing is unable to ignore that error because line numbers have changed in the cov-errors.txt.

zibaiwan (approved):
Hi Tyler, it looks good to me! Please squash the commit and I will merge.

zibaiwan:
Hi Tyler, can you please correct the commit message? Thanks

tylerzhao7684:
> Hi Tyler, can you please correct the commit message? Thanks

Updated

zibaiwan (approved):

zibaiwan:
Thanks Tyler!

PR #298, opened 2023-06-06
Fuzz Testing Implementation

tylerzhao7684:
- [x] Fuzz Testing Directory and contents added
- [x] CMake configured for Fuzz Testing directory
- [x] Header file created to allow unit test to load data from a properly-formatted yaml file
- [x] Python script created to mutate a single variable and output to a new file
- [x] Python script created to do fuzz testing for a single unit test file
- [x] Python script created to do fuzz testing for multiple test files
- [x] Save fuzz testing outputs to a new directory
- [x] Load data for different data types
- [x] Identify which variables are suitable for fuzzing
- [x] Workaround for unit test memory leak issue
- [x] Interpret results from fuzz testing and determine if a failure is "secure"
- [x] Code Coverage Report
- [x] Workflow that triggers fuzz testing
- [x] Documentation on how to do fuzz testing locally


zibaiwan:
Thanks Tyler for working on this. The approach looks good to me! I assume we won't check-in all unit tests file unless we have the corresponding yml file to fuzz test them, correct? I will take a closer look on the scripts too.

tylerzhao7684:
Proof that the workflow is correctly catching Address Sanitizer errors (ASAN errors):
[This workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5457221443/jobs/9930995686?pr=298#step:10:12) has 3 ASAN errors, which is due to [this change](https://github.com/intel/fpga-runtime-for-opencl/pull/298/commits/7388b447d19afaae2bcb18c8ecc6cc8c64dd4064), the reason why there are 3 is because that test is ran 3 times, due to [3 mutable variables](https://github.com/intel/fpga-runtime-for-opencl/pull/298/files#diff-de0de5bdf3bea48b50f4e3dd15eda4902df95e609405b26f43a4f78025e033fdR3-R5) in the input yml file

tylerzhao7684:
Proof that the test is able to detect hangs correctly:
[This workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5457395556/jobs/9931363071?pr=298#step:10:17) has 18 hangs because I changed the [timeout to 1 second](https://github.com/intel/fpga-runtime-for-opencl/pull/298/commits/05d95af295ab47edac3688b9b43c291d486f54b1#diff-c9879dbf7198c9e03e10ea98ec89c52d8c8775b045410c95c86db5f01f9878d1R51) which is obviously not enough. The point is that the script is able to detect hang correctly and I would expect it to do the same if a test does not finish in 60 seconds

tylerzhao7684:
Proof that the test is able to catch assertion failures correctly:
[This workflow run](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5457508306/jobs/9931604500?pr=298#step:10:15) has 4 assertion failures because of [this change](https://github.com/intel/fpga-runtime-for-opencl/pull/298/commits/0aa6d8037c7a12cab05cfc5aba058f42a62a60d4). There are 4 of them, 3 of them are because that the same test ran 3 times, due to [3 mutable variables](https://github.com/intel/fpga-runtime-for-opencl/pull/298/files#diff-de0de5bdf3bea48b50f4e3dd15eda4902df95e609405b26f43a4f78025e033fdR3-R5) in the input yml file. The remaining assertion failure is an actual assertion failure due to mutated variable (Download the artifact [here](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5457508306?pr=298) if you want to see the test output)

fuzz_testing/test/acl_globals_test.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2010-2021 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
```
zibaiwan, line 3: I only see `acl_auto_configure_test.cpp` is being tested in the workflow run. Are other tests being fuzz tested as well? If they are not fuzz tested, we should remove them. If you have made changes to it and it's for the future release, you can move it to another draft PR.

fuzz_testing/test/board/a10_ref/board_env.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0"?>
+<!--
```
zibaiwan, line 2: If the `board/a10_ref`, `hardware/a10_ref_small` and `fake_bsp` are exactly same as the one in the non fuzz test directory. I don't think we should checkin those files again. Can you see if you have anyway to reduce the redundancy here?

fuzz_testing/test/acl_auto_configure_test.cpp
```diff
@@ -0,0 +1,1752 @@
...
+#include <acl_version.h>
+
+#include "acl_test.h"
+
+#include "../fuzz_src/fuzz_testing.h"
+#include <vector>
+
+TEST_GROUP(auto_configure) {
+public:
+  void setup() { preload_data("acl_auto_configure_test"); }
```
zibaiwan, line 31: Can you please add some comments/documentation around the code that you added specially for the fuzz testing so it makes the reader easier to see what code is original and what is not?

zibaiwan (commented):
Hi Tyler, I have reviewed some of your changes. They look very good to me. I had some small comments below. I will continue reviewing some technical details and add more comments.

.github/workflows/fuzz-test.yml
```diff
@@ -0,0 +1,73 @@
...
+    inputs:
+      num_of_iterations:
+        description: Number of iterations per fuzzable variable
+        required: False
+        default: 5
+
+jobs:
+  build:
+    runs-on:
+      - self-hosted
```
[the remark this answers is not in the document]
tylerzhao7684: Not really, I copied the format from the coverity workflow. However I have configured the workflow to work on the self-hosted machine, is there any reason why I can't use the self-hosted machine?

fuzz_testing/test/acl_globals_test.h
tylerzhao7684: I believe the other test source code are included because at some point `acl_auto_configure_tet.cpp` called functions in other test cpp files. I have removed the other tests that `acl_auto_configure_tet.cpp` won't use.

tylerzhao7684:
Thanks for the review Zibai! I thought about it and I changed my mind on [this](https://github.com/intel/fpga-runtime-for-opencl/pull/298/files#diff-3ad568debbfef96b4adc342521a65a8f0d814a50e28b5608aef0ce45e8d9ec2aR42). I believe although I built the fuzz test infra based off the current unit tests, we should not regard them as the same test. In other words. we can add different contents to the fuzz tests and the unit tests. The commonality of fuzz tests and unit tests is that they test the same source function (i.e. `src/acl_auto_configure.cpp`). The difference is that one tests for security vulnerability, the other one tests for correctness. 

This opens up opportunity to create tests that are fuzz test only, such as a more in-depth test that goes through multiple source functions that interacts with the fuzzed variable.

I hope this answers:
> I assume if any of those auto_discovery string is updated in the test/acl_auto_configure_test.cpp, they need to be updated here?

> We update that file quite frequently, and then the developer needs to update this file and the yml file, which add more workload to the developer.

> If the board/a10_ref, hardware/a10_ref_small and fake_bsp are exactly same as the one in the non fuzz test directory. I don't think we should checkin those files again. Can you see if you have anyway to reduce the redundancy here?

What do you think?

zibaiwan:
> Thanks for the review Zibai! I thought about it and I changed my mind on [this](https://github.com/intel/fpga-runtime-for-opencl/pull/298/files#diff-3ad568debbfef96b4adc342521a65a8f0d814a50e28b5608aef0ce45e8d9ec2aR42). I believe although I built the fuzz test infra based off the current unit tests, we should not regard them as the same test. In other words. we can add different contents to the fuzz tests and the unit tests. The commonality of fuzz tests and unit tests is that they test the same source function (i.e. `src/acl_auto_configure.cpp`). The difference is that one tests for security vulnerability, the other one tests for correctness.
> 
> This opens up opportunity to create tests that are fuzz test only, such as a more in-depth test that goes through multiple source functions that interacts with the fuzzed variable.
> 
> I hope this answers:
> 
> > I assume if any of those auto_discovery string is updated in the test/acl_auto_configure_test.cpp, they need to be updated here?
> 
> > We update that file quite frequently, and then the developer needs to update this file and the yml file, which add more workload to the developer.
> 
> > If the board/a10_ref, hardware/a10_ref_small and fake_bsp are exactly same as the one in the non fuzz test directory. I don't think we should checkin those files again. Can you see if you have anyway to reduce the redundancy here?
> 
> What do you think?

Hi Tyler, yeah I agree we should consider those tests are different. But I think the developer should still update the fuzz test if they make the change to the acl_auto_configure.cpp, right? I guess you meant they are similar right now but they can be very different in the future, I agree with you and that means the automation is not neccessary.

However, I do believe we should not create duplicate files of those bsp files as I don't think we will do anything different for BSP in fuzz testing. Can you check if there is anyway the fuzz test can just reuse those BSPs?

tylerzhao7684:
> > Thanks for the review Zibai! I thought about it and I changed my mind on [this](https://github.com/intel/fpga-runtime-for-opencl/pull/298/files#diff-3ad568debbfef96b4adc342521a65a8f0d814a50e28b5608aef0ce45e8d9ec2aR42). I believe although I built the fuzz test infra based off the current unit tests, we should not regard them as the same test. In other words. we can add different contents to the fuzz tests and the unit tests. The commonality of fuzz tests and unit tests is that they test the same source function (i.e. `src/acl_auto_configure.cpp`). The difference is that one tests for security vulnerability, the other one tests for correctness.
> > This opens up opportunity to create tests that are fuzz test only, such as a more in-depth test that goes through multiple source functions that interacts with the fuzzed variable.
> > I hope this answers:
> > > I assume if any of those auto_discovery string is updated in the test/acl_auto_configure_test.cpp, they need to be updated here?
> > 
> > 
> > > We update that file quite frequently, and then the developer needs to update this file and the yml file, which add more workload to the developer.
> > 
> > 
> > > If the board/a10_ref, hardware/a10_ref_small and fake_bsp are exactly same as the one in the non fuzz test directory. I don't think we should checkin those files again. Can you see if you have anyway to reduce the redundancy here?
> > 
> > 
> > What do you think?
> 
> Hi Tyler, yeah I agree we should consider those tests are different. But I think the developer should still update the fuzz test if they make the change to the acl_auto_configure.cpp, right? I guess you meant they are similar right now but they can be very different in the future, I agree with you and that means the automation is not neccessary.
> 
> However, I do believe we should not create duplicate files of those bsp files as I don't think we will do anything different for BSP in fuzz testing. Can you check if there is anyway the fuzz test can just reuse those BSPs?

Yup I'll remove the BSPs in the fuzz_testing directory. 
Note that we only reference BSP when doing `export AOCL_BOARD_PACKAGE_ROOT="$(git rev-parse --show-toplevel)/test/board/a10_ref"` as documented in runtime README, as a pre-requisite for running unit tests.

.github/workflows/fuzz-test.yml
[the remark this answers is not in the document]
tylerzhao7684: > Ideally we want to get rid of / minimize all self-hosted machine to reduce maintainer cost. Currently only Coverity requires self-hosted machine due to it's not available (license) in the github machines.

I tried using the github hosted machine but [it failed with ssl certificate error](https://github.com/intel/fpga-runtime-for-opencl/actions/runs/5592330451/jobs/10224649829#step:7:13) :(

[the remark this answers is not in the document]
zibaiwan: @tylerzhao7684, I found this github [page ](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python)about how to run python install. Your code looked fine. I think the only difference might be the image you are using is a customized image that we built for the runtime [here](https://github.com/intel/fpga-runtime-for-opencl/tree/main/container).

We also have a workflow to update all the container [here](https://github.com/intel/fpga-runtime-for-opencl/blob/main/.github/workflows/container.yml). Be careful with this `container.yml` as this will update all the containers with the possibility of polluting them. You can read the comment inside to see how to just update one container.

I would suggest you to try the code in that github tutorial to see if you can get python install to run properly using the github runner + github default image. If you can get that to run, maybe something is missing in our container. Before you start to update our container, you can create a new container for your testing (so you don't have the chance to break other containers). I suspect might be some SSL module needs to be installed.

However, this is not the top priority for this task. We can merge this in as the self-hosted machine and fix it before the CC.

fuzz_testing/test/acl_auto_configure_test.cpp
tylerzhao7684: I have added more documentation at the top of the acl_auto_configure_test.cpp

.github/workflows/fuzz-test.yml
```diff
@@ -0,0 +1,73 @@
+name: Runtime Fuzz Testing
+
+on:
+  # For the purpose of testing the workflow
+  pull_request:
```
zibaiwan, line 5: Todo: remove

fuzz_testing/fuzz_src/fuzz_testing.h
```diff
@@ -0,0 +1,196 @@
...
+    exit(1);
+  }
+}
+
+// Utility functions
+inline bool check_condition(bool condition, bool &check) {
+  check = condition;
+  return check;
+}
+#endif
```
zibaiwan, line 196: new line

zibaiwan (dismissed):
Thanks Tyler. I think this PR is ready to be merged once the `pull_request` testing removed in the yml.

zibaiwan (approved):

PR #300, opened 2023-07-07
Updated acl_context_callback With Additional Log

sharmag1-Gaurav:
With this change, supplying the environment variable ACL_CONTEXT_CALLBACK_DEBUG=1 will make the runtime output the errInfo stored in the acl_context_callback function. 

Minor refactoring and moving env strings to header file.

### DIFF
--- a/include/acl_context.h
+++ b/include/acl_context.h
@@ -16,6 +16,22 @@ extern "C" {
 #pragma GCC visibility push(protected)
 #endif
 
+// Constants
+static constexpr const char *ENV_CL_CONTEXT_ALLOW_MULTIPROCESSING_INTELFPGA =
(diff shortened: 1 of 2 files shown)


zibaiwan (commented):
Hi Gaurav,

Thanks for adding the additional log. The change looks good to me. Please squash the change into 1 commit since the Runtime repo merges multiple commits into the repo to allow one PR to fix multiple issues. In this case, one commit is more suitable since the clang-format is just a fixup.

I will merge once the commit is squashed. Another thing is, I think this will require a doc change in our programming guide, please track that later.

Best regards,

Zibai

zibaiwan (approved):

zibaiwan:
Thanks Gaurav!

PR #301, opened 2023-07-12
Add workaround for host pipe atomic hang issue

sophimao:
This is a workaround for the same issue as described in PR #270.


zibaiwan (approved):
Thanks Sophie for providing the workaround!

PR #307, opened 2023-08-11
Increase fuzz testing timeout to 10 minutes

tylerzhao7684:
In some cases, it does take up to 5 minutes to complete a fuzz test (with mutated input), therefore increasing timeout to 10 minutes


zibaiwan (approved):
Thanks Tyler

PR #308, opened 2023-08-11
Clean up runtime code for simulation preprogram autodiscovery string load

sophimao:
Unlike in hardware, in simulation the runtime is not able to obtain autodiscovery string
from the device ROM memory before reprogramming, therefore, global memory
configuration information is only available after device reprogram, during which the
autodiscovery string stored in the binary will be extracted. Currently the runtime does
some special handling for the simulation flow to amend for this behavioural difference.

With the newest change to our simulator, the autodiscovery string will be available for read
preprogram and the runtime can now obtain this information just as in the hardware flow
(except for some minor difference). Therefore, this change cleans up the runtime by removing
code for special handling of the simulation flow.


src/acl_kernel_if.cpp
```diff
@@ -846,8 +809,13 @@ int acl_kernel_if_init(acl_kernel_if *kern, acl_bsp_io bsp_io,
     return -1;
   }
 
-  result = acl_kernel_if_update(
-      sysdef->device[kern->physical_device_id].autodiscovery_def, kern);
+  if (acl_platform.offline_mode == ACL_CONTEXT_MPSIM) {
+    sysdef->device[kern->physical_device_id].autodiscovery_def.name =
```
zibaiwan, line 813: It seems like simulation still does some special handling here. Can you please add some comment to explain why the simulation cannot just call `acl_kernel_if_update` here?

zibaiwan:
Hi Sophie,

Thanks for the change. It looks good to me! I think you only ran SYCL tests for the change. Do you mind running a few opencl simulation tests to make sure your change doesn't break OpenCL flow?

src/acl_kernel_if.cpp
sophimao: Hmm I think it has something to do with `simulation_set_kernel_cra_address_map` but I can't remember what exactly it was... Let me check again

include/acl_hal_mmd.h
```diff
@@ -10,7 +10,112 @@
...
+// Versioning constants, address maps, and bit/byte positionings used   //
+// in acl_kernel_if and acl_pll                                         //
+//                                                                      //
+//////////////////////////////////////////////////////////////////////////
+
+//// acl_kernel_if
+// Versioning: This value must be read from addr 0
+// For unit tests to work, this defines must match the one in the unit test
+// header file
+#define KERNEL_VERSION_ID (0xa0c00001)
```
zibaiwan, line 22: It's great that those Macros are moved to a header file. I checked the original check-in of those values and couldn't see comments why they were put in the acl_kernel_if.c in the first place. I think the only thing needs to be cautious is to make sure those values are not defined duplicate in the compiler side as this header is included by the simulator.

sophimao: Yes I have removed the definition of these macros in the simulation library :)

src/acl_kernel_if.cpp
sophimao: Added more logic in the simulation runtime and removed the special handling, now we just set device name and num_devices for sim.

src/acl_hostch.cpp
```diff
@@ -41,7 +41,7 @@ static cl_int l_push_packet(unsigned int physical_device_id, int channel_handle,
   // Pushed data can be smaller than the request write_size due to compiler
   // padding. Runtime needs to check if the trailing bytes are all 0s.
   else if ((pushed_data > 0) && (pushed_data < write_size)) {
-    for (int i = pushed_data; i < write_size; i++) {
+    for (size_t i = pushed_data; i < write_size; i++) {
       unsigned char c = ((char *)host_buffer)[i];
       if (c != 0) {
         // This shouldn't happen. Needs to send out a warning to user rather
```
zibaiwan, line 47: Thanks for cleaning this up Sophie. Could you please modify the code a bit more?

```
        std::cerr << "Error: Data is not fully written into the Hostpipe. "
                     "None-0 bits have been cut off \n";
        assert(0);
```

I wanted to add an assert here since I realize the user would see this error repeatly if they are using a block read, so for now I think we can just force the program to stop if they fall into this path.

zibaiwan (dismissed):
Thanks Sophie. Looks good to me!

sophimao:
Resolved review comment and squashed commits, @zibaiwan would you mind reviewing again? Thanks!

zibaiwan (approved):
Thanks Sophie! Looks good to me !

PR #312, opened 2023-08-22
hold on to new commands

sharmag1-Gaurav:
if acl_submit_command doesn't work on a new command in the command queue, we should hold on to it because it could be due to the device op queue being full or some other reason. By holding on, we can try to submit again when there is more space in the device op queue

TODO: Testing

### DIFF
--- a/src/acl_command_queue.cpp
+++ b/src/acl_command_queue.cpp
@@ -630,20 +630,19 @@ int acl_update_queue(cl_command_queue command_queue) {
   }
 }
 
-int acl_try_FastKernelRelaunch_ooo_queue_event_dependents(cl_event parent) {
-  int num_updates = 0;
+void acl_try_FastKernelRelaunch_ooo_queue_event_dependents(cl_event parent) {
 
   if (!(parent->command_queue->properties &
         CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE))
-    return 0;
+    return;
   if (parent->depend_on_me.empty())
-    return 0;
+    return;
   if (parent->cmd.type != CL_COMMAND_TASK &&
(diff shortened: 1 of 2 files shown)


src/acl_command_queue.cpp
```diff
@@ -677,11 +677,13 @@ void acl_try_FastKernelRelaunch_ooo_queue_event_dependents(cl_event parent) {
...
-    dependent_it--; // decrement it other wise we will skip an element
-    dependent->command_queue->num_commands_submitted++;
-    acl_submit_command(dependent);
+    int local_updates = acl_submit_command(dependent);
+    if (local_updates) {
+      dependent->depend_on.erase(parent);
+      dependent_it = parent->depend_on_me.erase(dependent_it);
+      dependent_it--; // decrement it otherwise we will skip an element
+      dependent->command_queue->num_commands_submitted++;
+    }
```
sophimao, line 686: Just curious, how does not returning num_update resolve the unit test hang?

sophimao (approved):
Thanks Gaurav, this looks good!

src/acl_command_queue.cpp
sharmag1-Gaurav: Talked to Sophie about this over a call. Fast Kernel Relaunch, is a feature that's okay to fail. In the previous version of the change, it would consider FKR failing as the new command failing to be submitted. However, it's okay if the new command fails to be submitted by FKR, since it'll be picked up again and submitted when its parent command (the one it depends on, which must exist since FKR happens in this case only) finishes.

sharmag1-Gaurav:
@sophimao could you merge this in when you get the chance? I don't think I have access. Realistically, this should prob be done Tuesday

PR #318, opened 2023-09-28
Update invocation image debug prints

sophimao:
The runtime prints invocation image data in steps size of integer size, this doesn't take into the account when the last block of data has a smaller size. This change is added to handle this special case.


zibaiwan (approved):
Thanks Sophie!

PR #319, opened 2023-09-29
Remove simulator device from offline devices

sophimao:
Simulator device will now be initialized in the same fashion as hardware devices,
during `l_initialize_devices`. This can be done as we now load simulator device
global memory configuration from the correct autodiscovery string generated by
the system integrator as introduced by commit https://github.com/intel/fpga-runtime-for-opencl/commit/d57cedc9fa35d2efba22892e0d3e715481db09d3.


zibaiwan (approved):
Thanks Sophie! Looks good to me.

PR #322, opened 2023-10-18
Delay erasing context from contexts_set during `acl_idle_update`

sophimao:
Fixes issue where context is being deleted from `contexts_set` while traversing through the same `context_set`, causing undefined behaviour and sometimes crash in `acl_idle_update`.


zibaiwan (dismissed):
Thanks Sophie. Looks good except a minor comment.

zibaiwan (approved):

PR #324, opened 2023-11-15
Only write changing parts of kernel arguments to kernel CRA

sophimao:


zibaiwan:
Thanks Sophie. The overall PR looks good to me. Just one comment regarding using C code vs C++ code. I think since the Runtime have migrated to C++, we probably want to adopt C++ features in the case. Like using new&delete vs malloc&free, it's recommended to never use malloc in C++ due to a number of reasons.

For example

Instead of using 
`kern->kernel_arg_cache = (char **)acl_malloc(kern->num_accel * sizeof(char *));`

Maybe use

`kern->kernel_arg_cache ** db = new char*[kern->num_accel]();` Note: The () at the end should initialize each of the pointer to nullptr for you.

Don't forget to use `delete []` for the array and `delete` each pointer.

While I am typing those, I feel I don't want to handle memory management myself already. I feel it's better to use the modern  C++ feature `unique_ptr` and `vector` here as the underlying type of `kernel_arg_cache`.

I will let you decide.

zibaiwan (dismissed):

sophimao:
Thanks Zibai for the suggestion! I think it's definitely valid, but I might submit that as a separate change in another PR as there are a lot of members of `acl_kernel_if` that currently still use `malloc`. In the separate change I might just update everything at once.

zibaiwan (approved):

PR #325, opened 2023-11-17
Use c++ constructs to replace malloc calls in acl_kernel_if

sophimao:


src/acl_kernel_if.cpp
```diff
@@ -1187,22 +1165,22 @@ void acl_kernel_if_launch_kernel_on_custom_sof(
...
       (kern->csr_version != CSR_VERSION_ID_18_1 && image->arg_value_size > 0)) {
-    if (kern->accel_arg_cache[accel_id] == nullptr) {
+    if (!kern->accel_arg_cache[accel_id]) {
       acl_kernel_cra_write_block(
           kern, accel_id, offset + (unsigned int)image_size_static,
           (unsigned int *)image->arg_value, image->arg_value_size);
-      kern->accel_arg_cache[accel_id] =
-          (char *)acl_malloc(image->arg_value_size);
-      memcpy(kern->accel_arg_cache[accel_id], (char *)image->arg_value,
+      kern->accel_arg_cache[accel_id].reset(new char[image->arg_value_size]);
```
zibaiwan, line 1172: Thanks Sophie.

Nitpick, I think it's better to use `make_unique` instead of `new` here. I think the main point of leveraging unique pointer so that we don't have to call `new` or `delete` ourselves among other reasons why `make_unique` is better than `new`.

Everything else looks good to me.

PR #326, opened 2023-11-27
Fix csr pipe write handshaking

zibaiwan:
1. If blocking, read valid reg, wait until valid is 0.
2. If non-blocking, read valid reg once ->return failure if valid is 1.
3. write to the pipe.
4. write 1 to the valid.

### DIFF
--- a/src/acl_hostch.cpp
+++ b/src/acl_hostch.cpp
@@ -966,6 +966,13 @@ void acl_write_program_hostpipe(void *user_data, acl_device_op_t *op) {
 
   if (host_pipe_info.implement_in_csr) {
     // Get CSR address
+
+    // Here is the logic for CSR pipe write
+    // 1. If blocking, read valid reg, wait until valid is 0.
+    // 2. If non-blocking, read valid reg once ->return failure if valid is 1.
+    // 3. write to the pipe.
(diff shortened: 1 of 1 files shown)


src/acl_hostch.cpp
```diff
@@ -991,24 +1021,14 @@ void acl_write_program_hostpipe(void *user_data, acl_device_op_t *op) {
...
-    // In non-blocking case, there is no need to write into valid register.
-    // We only care about valid register if the protocol is
-    // avalon_mm_uses_ready.
-    if (blocking && host_pipe_info.protocol == 3) {
-      // Tell CSR it's valid
-      acl_get_hal()->write_csr(host_pipe_info.m_physical_device_id, valid_reg,
-                               (void *)&valid, (size_t)sizeof(uintptr_t));
+    // For now, we trust the AVALON_MM by default uses valid. TODO: fix this
+    // later by use the new protocol info Provided by the compiler Tell CSR it's
+    // valid.
```
sophimao, line 1026: Nit: split clang-formatted comment into multiple lines

sophimao (approved):
Thanks Zibai! LGTM except for a minor nit on comment.

sophimao (approved):
Thanks Zibai!

PR #330, opened 2023-12-13
Pass information on whether a streaming kernel has CRA arguments to MMD

sophimao:


src/acl_kernel_if.cpp
```diff
@@ -1163,6 +1163,7 @@ void acl_kernel_if_launch_kernel_on_custom_sof(
                                image_size_static);
   }
 
+  bool accel_has_agent_args = false;
```
zibaiwan, line 1166: Hi Sophie,

Silly question, what's the point of passing this parameter if the value is always false?

sophimao: Not silly at all! There was some merging/reverting happening in my workspace and apparently I must have somehow deleted the line that modified it, I'll add it back and rerun the tests.

zibaiwan (approved):
Thanks Sophie!

PR #334, opened 2023-12-18
Runtime clean up

sophimao:


src/acl_kernel_if.cpp
```diff
@@ -817,10 +817,8 @@ int acl_kernel_if_update(const acl_device_def_autodiscovery_t &devdef,
           devdef.accel[ii].iface.name.c_str(), kern->accel_csr[ii].address,
           kern->accel_csr[ii].bytes);
     }
-    if (acl_platform.offline_mode == ACL_CONTEXT_MPSIM) {
```
zibaiwan, line 820: What the benefit of moving this if statement inside of the `acl_hal_mmd_simulation_set_kernel_cra_address_map` function? It's kind of strange that always calling this function but it just does nothing when the mode is not simulation though.

include/acl.h
```diff
@@ -593,20 +593,19 @@ typedef struct acl_device_def_autodiscovery_t {
 } acl_device_def_autodiscovery_t;
 
 typedef struct acl_device_def_t {
-  // Back pointer. This is only valid if this def is owned by a dev_prog. In
-  // tests, this is not the case.
-  acl_device_program_info dev_prog;
```
zibaiwan, line 598: Do you know why this point exists in the first place? So from now on, dev_prog only lives in the `acl_device_binary_t` data structure now, correct?

zibaiwan (commented):
Thanks Sophie. Looks good to me. I just had a couple questions mainly for my own knowledge.

src/acl_kernel_if.cpp
sophimao: Good question! As per Joe's suggestion the runtime should be agnostic to whether we are running in simulation or not, so this is a preparation step for the removal of the ACL_CONTEXT_MPSIM mode altogether.

include/acl.h
sophimao: > Do you know why this point exists in the first place?

I might know at a point of time but I don't remember anymore :p

> So from now on, dev_prog only lives in the acl_device_binary_t data structure now, correct?

Based on my notes looks like the true copies of `acl_device_program_info` belongs to a `cl_program`, and the dev_prog in `acl_device_binary_t` is still just a back-pointer, but unlike the used back-pointer being removed in this change, it is used in acl_device_binary.cpp to get device, etc.

zibaiwan (approved):

PR #335, opened 2023-12-19
acl_hal_mmd clean up: purely cosmetic change

sophimao:


sophimao:
Commits shall be merged post-review

zibaiwan:
> Commits shall be merged post-review

Thanks Sophie for clean this up. I actually like the separate commits (you can merge the clang-format one into one of the commit). It was very clear to understand what you are changing only by reading the separate commits.

zibaiwan (approved):
Thanks Sophie. I will re-approve if you want to merge the clang-format one into one of the commit.

sophimao:
> you can merge the clang-format one into one of the commit

Ah okay, one thing with the commits is that I think the first commit doesn't work on its own as I forgot to test in between commit 1 and commit 2 😢 Would it be okay if I combine commits 1 and 2, and leave commit 3+clang-format as a separate commit?

zibaiwan:
> > you can merge the clang-format one into one of the commit
> 
> Ah okay, one thing with the commits is that I think the first commit doesn't compile on its own as I forgot to test in between commit 1 and commit 2 😢 Would it be okay if I combine commits 1 and 2, and leave commit 3+clang-format as a separate commit?

Sure, sounds good!

PR #338, opened 2024-01-03
Added top-level token permission and move write permission to job level for security reason

zibaiwan:
Testing is done from my personal fork with temporarily modified the registry.


sophimao (approved):
Thanks Zibai!

PR #341, opened 2024-02-22
Prerequisite changes for getting rid of simulation env var

sophimao:


src/acl_kernel_if.cpp
```diff
@@ -716,15 +716,6 @@ int acl_kernel_if_init(acl_kernel_if *kern, acl_bsp_io bsp_io,
   auto load_result = acl_load_device_def_from_str(
       config_string, sysdef->device[kern->physical_device_id].autodiscovery_def,
       auto_config_err_str);
-  if (load_result) {
-    if (acl_platform.offline_mode == ACL_CONTEXT_MPSIM) {
-      sysdef->device[kern->physical_device_id].autodiscovery_def.name =
-          ACL_MPSIM_DEVICE_NAME;
-      sysdef->num_devices = 1;
-    } else {
-      ++sysdef->num_devices;
```
zibaiwan, line 725: Quick question: what is the reason that sysdef->num_devices doesn't need to be updated anymore in this function?

zibaiwan:
Thanks Sophie. Overall looks good to me except one question that I have.

src/acl_kernel_if.cpp
sophimao: Based on my understanding `sysdef->num_devices` is initiated by probing each loaded MMD with the number of devices they have when platform gets initialized, and is used once when calculating the number of devices attached to the platform during the platform initialization.

After the platform is initialized and during the context creation time, `sysdef->num_devices` is only used to print debug messages, so I think we don't need to increment it when `l_try_device` is called, otherwise in sycl where multiple `clGetDeviceInfo` is called,`sysdef->num_devices` will gets incremented to way more than the actual number of devices enumerated by the MMD.

zibaiwan: Thanks Sophie. So basically `sysdef->num_devices` is set by `acl_mmd_get_system_definition()` and it was a mistake to update the value during the `l_try_device()`.

zibaiwan (approved):

PR #342, opened 2024-02-27
Update MMD versioning scheme

### LINKED ISSUE #147: Query whether the MMD supports buffer_location in shared/host USM allocations API calls

sophimao:
Resolves #147 by deciding buffer location support based on the MMD version.

With the new versioning scheme, MMDs starting from version 2024.2 should be supporting buffer location and accepting the buffer location property.


src/acl_usm.cpp
```diff
@@ -143,12 +143,14 @@ CL_API_ENTRY void *CL_API_CALL clHostMemAllocINTEL(
...
   }
 
+  bool track_mem_id = false;
   if (acl_get_hal()->host_alloc) {
     std::array<mem_properties_t, 3> mmd_properties;
     {
       auto mmd_properties_it = mmd_properties.begin();
       if (mem_id) {
-        if (acl_platform.offline_mode == ACL_CONTEXT_MPSIM) {
+        if (acl_get_hal()->support_buffer_location(devices)) {
```
zibaiwan, line 152: Will this break backward compatibility on simulation device prior to 2024.2?

zibaiwan: Note: since BL is a SYCL/oneAPI feature instead of the OpenCL feature and we don't have hard backward compatibility requirement for oneAPI. If there is no way to keep the backward compatibility of this, I think this is OK to break it.

sophimao: > Will this break backward compatibility on simulation device prior to 2024.2? E.g., new runtime will not work with the 2024.1 compiled exe for buffer locations.

Yes it will break backwards compatibility, I talked with Joe and the expectation is that users will use compatible simulation MMD and the runtime since they are bundled together in oneAPI.

zibaiwan: Sounds good!

src/acl_hal_mmd.cpp
```diff
@@ -965,22 +996,40 @@ void l_get_physical_devices(acl_mmd_dispatch_t *mmd_dispatch,
   mmd_dispatch->aocl_mmd_get_offline_info(AOCL_MMD_VERSION, sizeof(buf), buf,
                                           NULL);
   buf[sizeof(buf) - 1] = 0;
-  mmd_dispatch->mmd_version = atof(buf);
+  mmd_dispatch->mmd_version = l_parse_mmd_version_str(std::string(buf));
+  if (mmd_dispatch->mmd_version < 0) {
+    printf("  Invalid MMD version:     %s\n", buf);
+    printf("Contact the board package support vendor for resolution.\n");
```
sophimao, line 1002: @aktole Could you check if the printed message here is okay?

src/acl_hal_mmd.cpp
```diff
@@ -965,22 +996,40 @@ void l_get_physical_devices(acl_mmd_dispatch_t *mmd_dispatch,
...
+  if (MMDVERSION_LESSTHAN(hal_version,
                           mmd_dispatch->mmd_version) || // MMD newer than HAL
       MMDVERSION_LESSTHAN(mmd_dispatch->mmd_version,
                           14.0)) // Before this wasn't forward compatible
   {
     printf("  Runtime version: %s\n", AOCL_MMD_VERSION_STRING);
     printf("  MMD version:     %s\n", buf);
+    printf("MMD version is newer than the runtime version! Use the runtime "
+           "with version greater or equal to the MMD version, or contact the "
+           "board support package vendors if mismatch is unexpected.\n");
```
sophimao, line 1030: @aktole Could you check if the printed message here is okay?

src/acl_hal_mmd.cpp
aktole: Yes, looks good. This condition would probably never be true, unless BSP vendor sets the MMD version to something less than 0 accidently right?

src/acl_hal_mmd.cpp
aktole: Yes, looks good, thanks Sophie

aktole (approved):

sophimao:
> Yes, looks good. This condition would probably never be true, unless BSP vendor sets the MMD version to something less than 0 accidently right?

Right, or unless they set it to some jibber-jabber that cannot be converted to a floating point number, which I don't think will happen very often :p

zibaiwan:
Hi Sophie, 

Is this planned to go to 2024.2 release?

Thanks,

Zibai

zibaiwan:
> Hi Sophie,
> 
> Is this planned to go to 2024.2 release?
> 
> Thanks,
> 
> Zibai

Chatted with Sophie offline. This will go to Trunk first and then cherry-picked into 2024.2 depends on the full test results. I am reviewing the change, so far looks good to me. Will do a more thorough review and should be done by tomorrow.

zibaiwan (approved):
Thanks Sophie!

zibaiwan (approved):

PR #343, opened 2024-02-29
Support arbitary BL for accessor

zibaiwan:
Tested with a variety of tests group.

### DIFF
--- a/include/acl.h
+++ b/include/acl.h
@@ -381,6 +381,8 @@ typedef struct {
 
     return result;
   }
+
+  std::string id = "-";
 } acl_system_global_mem_def_t;
 
 // Our allocator is optimized for simplicity, and for the case where there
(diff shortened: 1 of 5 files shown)


include/acl.h
```diff
@@ -381,6 +381,8 @@ typedef struct {
 
     return result;
   }
+
```
sophimao, line 384: Maybe put it before the function so that it doesn't get missed when reading through the code?

sophimao (approved):
Thanks Zibai! LGTM, just one small nit below that you can make the call whether you want to change it or not :)

include/acl.h
zibaiwan: Thanks Sophie for reviewing. I think I will leave it as is for now. I tend to always add new data member to the existing struct at the end of struct for avoiding aggregate initialization of the struct somewhere else in the code base (not sure if it's a good/bad practice). Although in this case, put it before the function should be OK as it's still the end of the grouped data members. Since you don't seem to have strong opinion on this, I will leave it for now.

zibaiwan:
I made a minor update here in the auto_configure.cpp

`std::string gmem_id = "-"; // This should be the default value even if the
                               // auto-discovery string doesn't have this field.`

sophimao (approved):
Thanks Zibai!

PR #349, opened 2024-03-27
Set default values for kernel image static part and skip CSR write if no change

sophimao:


zibaiwan (dismissed):
Thanks Sophie! Looks good to me!

zibaiwan (approved):

PR #351, opened 2024-04-08
Update loaded_bin if reprogram is skipped at kernel launch

sophimao:


zibaiwan (approved):
Thanks Sophie! Looks good to me!

PR #360, opened 2024-05-01
Fix USM mem blocking free corruption

sophimao:
If the event being waited on is completed and deleted during the traversal of `command_queue->{commands|inorder_commands}`, this would lead to corruption of the corresponding data structure. This change fixes this issue by deferring the deletion till the end of the current iteration.


sophimao:
Similar issue as #322

rho180 (approved):

PR #362, opened 2024-06-04
Added glibc_wrap to provide Linux OS backwards compatibility

jackzhaolt:
Our CI/CD is moving to SUSE15, but we need to support a GLIBC version (2.28) that is below the native GLIBC version on SUSE15) 
This change implements a wrapping mechanism that will explicitly link against the older GLIBC functions or syscall functions to the extent required for our oldest supported GLIBC (2.28)

The implementation is based on that used in hld-backend and DSPBA https://github.com/intel-innersource/applications.fpga.oneapi.products.hld-backend-source/pull/510

### Tests run:
I've verified with nm that without linking the wrapper, the `libalteracl.so` file uses GLIBC 2.29 for `log` operation and GLIBC 2.30 for `gettid`. With the wrapper it uses glibc2.2.5 for log operation and syscall for `gettid`.

Local build also passes with this change: https://psg-sc-arc.sc.intel.com/arc/dashboard/reports/show_job/941516093

### DIFF
--- a/glibc_wrap/glibc_wrap.c
+++ b/glibc_wrap/glibc_wrap.c
@@ -1,0 +1,30 @@
+// These functions are only called by our code when built on newer distros. We
+// add '-Wl,--wrap={funcname}' to the link line which changes all calls to
+// '{funcname}' to '__wrap_{funcname}'
+
+// For all the gritty details:
+// https://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html
+// https://stackoverflow.com/questions/4032373/linking-against-an-old-version-of-libc-to-provide-greater-application-coverage
+
+#include <linux/unistd.h>
+#include <math.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#ifndef __arm__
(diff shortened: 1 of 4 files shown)


smffraser (approved):

glibc_wrap/glibc_wrap.c
```diff
@@ -0,0 +1,33 @@
+// Derived from the version used in Quartus by Devin Sundaram and Jeff Da Silva:
+// (https://github.com/intel-innersource/applications.fpga.quartus-prime.glibc-wrap).
```
sophimao, line 2: It would be good to try avoid references to internal links since the runtime repo is a public one, and this link will appear as invalid address to users.

sophimao (commented):
Thanks Jack, could you also squash the commits since the runtime repo doesn't automatically squash them when merging to the main branch? You should be able to do it through
```
git rebase -i main # squash the commits into a single one
git push <your_remote> --force-with-lease
```

sophimao:
CI failure due to https://github.com/actions/runner-images/issues/10004, will bypass CI and merge

PR #363, opened 2024-06-04
Use mutex to ensure segment update and CRA read/write happens atomically

sophimao:
When multiple threads may access the kernel CRA registers and there are enough kernel that segment register may be modified throughout the run, we need to ensure there is no more additional segment updates coming in between a segment update and consequent CRA read/write.

### DIFF
--- a/src/acl_kernel_if.cpp
+++ b/src/acl_kernel_if.cpp
@@ -400,20 +400,42 @@ static uintptr_t acl_kernel_cra_set_segment_rom(acl_kernel_if *kern,
 
 static int acl_kernel_cra_read(acl_kernel_if *kern, unsigned int accel_id,
                                unsigned int addr, unsigned int *val) {
+  int result = 0;
   assert(kern->cra_ring_root_exist);
-  uintptr_t segment_offset = acl_kernel_cra_set_segment(kern, accel_id, addr);
-  acl_assert_locked_or_sig();
-  return acl_kernel_if_read_32b(
-      kern, (unsigned)OFFSET_KERNEL_CRA + (unsigned)segment_offset, val);
+#ifdef __linux__
+  acl_sig_block_signals();
+#endif
+  {
+    std::lock_guard<std::mutex> lock(kern->segment_mutex);
+    uintptr_t segment_offset = acl_kernel_cra_set_segment(kern, accel_id, addr);
+    acl_assert_locked_or_sig();
+    result = acl_kernel_if_read_32b(
+        kern, (unsigned)OFFSET_KERNEL_CRA + (unsigned)segment_offset, val);
+  }
+#ifdef __linux__
+  acl_sig_unblock_signals();
+#endif
+  return result;
 }
 
 int acl_kernel_cra_read_64b(acl_kernel_if *kern, unsigned int accel_id,
                             unsigned int addr, uint64_t *val) {
+  int result = 0;
   assert(kern->cra_ring_root_exist);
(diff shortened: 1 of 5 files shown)


justin-rosner (commented):
Are there any unit tests for the runtime?

sophimao:
> Are there any unit tests for the runtime?

Yes, they are run automatically by the CI

src/acl_kernel_if.cpp
```diff
@@ -400,20 +400,42 @@ static uintptr_t acl_kernel_cra_set_segment_rom(acl_kernel_if *kern,
...
+  acl_sig_block_signals();
+#endif
+  {
+    std::lock_guard<std::mutex> lock(kern->segment_mutex);
+    uintptr_t segment_offset = acl_kernel_cra_set_segment(kern, accel_id, addr);
+    acl_assert_locked_or_sig();
+    result = acl_kernel_if_read_32b(
+        kern, (unsigned)OFFSET_KERNEL_CRA + (unsigned)segment_offset, val);
+  }
+#ifdef __linux__
```
mendell27, line 415: Just a general comment.  It would be preferable to have a class whose constructor blocks the signals and whose dtor unblocks them.  Then you don't have to add expliciit code before every return (and would also handle exceptions.
You can make the ctor/dtor only actually do something on Linux, and then you can add it without an #ifdef to a place like this

You also wouldn't need to add a dummy result

src/acl_kernel_if.cpp
```diff
@@ -400,20 +400,42 @@ static uintptr_t acl_kernel_cra_set_segment_rom(acl_kernel_if *kern,
...
+  int result = 0;
   assert(kern->cra_ring_root_exist);
-  uintptr_t segment_offset = acl_kernel_cra_set_segment(kern, accel_id, addr);
-  acl_assert_locked_or_sig();
-  return acl_kernel_if_read_64b(
-      kern, (unsigned)OFFSET_KERNEL_CRA + (unsigned)segment_offset, val);
+#ifdef __linux__
+  acl_sig_block_signals();
+#endif
+  {
```
mendell27, line 428: Is there a reason why you need to put this into a new scope?  Does the mutex need to be release before the unlock?  If you do as I suggested, that would be done as well

sophimao: The new scope is for the mutex to unlock when exiting the scope, but yea I'll try the change you suggested above, maybe that way I can avoid this new scope.

src/acl_kernel_if.cpp
sophimao: Okay I created a class for linux signal blocking, I guess I could potentially pass the mutex to the ctor and dtor for locking/unlocking, but I'm not sure if this is making the blocker a little bit "too smart" since there are other place (acl_hal_mmd) where the blockers are used without the mutex?

src/acl_kernel_if.cpp
mendell27: You can remove the scope.  C++ rules say that the mutex will be released before the unblock of the signals

sophimao: Ah is it because of this?
https://stackoverflow.com/questions/2254263/order-of-member-constructor-and-destructor-calls

mendell27: yes.  they will be destructed in reverse order from the construction

sophimao: Scope removed

mendell27 (approved):

justin-rosner (approved):

PR #372, opened 2024-07-31
Handle printf format as the last item in the format string

mendell27:
The code in l_dump_printf_buffer incorrectly assumes that each printf %format in a format string is followed by at least one character.  In the case of **printf("%d", 10);**, this is not the case, and the code will not print the value.

The fix is to exit only when the format string has been processed and there is no conversion to be done.
Note that most printf formats end with an **\n**, so this problem doesn't occur.


sophimao (approved):
LGTM

